PR #4, opened 2025-11-17
Mitigate judge's order bias with swap mode

Ali-Elganzory:


openjury/generate_and_evaluate.py
```diff
@@ -89,6 +108,7 @@ def parse_args(cls):
             judge_model=args.judge_model,
             n_instructions=args.n_instructions,
             provide_explanation=args.provide_explanation,
+            swap_mode=args.swap_mode,
```
geoalgo: At this point args.swap_mode is a string but we map it to a boolean, we should probably rather test `args.swap_mode == "fixed"` rather?

openjury/generate_and_evaluate.py
```diff
@@ -244,6 +289,15 @@ def main(args: CliArgs):
...
     )
 
+    if args.swap_mode:
+        prefs_reversed = pd.Series(
+            [
+                score_parser.parse_model_raw(annotation.judge_completion)
+                for annotation in annotations_reversed
+            ]
+        )
+        prefs = (prefs + (1 - prefs_reversed)) / 2.0
```
geoalgo: Does it work in case of nans? Seems like return `2n` battles would be cleaner?

openjury/generate_and_evaluate.py
Ali-Elganzory: Yes, you are right. This got lost in the change from `correct_order_bias` (bool) to `swap_mode` (str).

Ali-Elganzory:
I have created #6

PR #9, opened 2026-02-14
Add documentation for LlamaCpp

ErlisLushtaku:
# Description

- Improve documentation for the usage of LlamaCpp.
- Added optional dependency on `llama-cpp-python>=0.3.0`

## Other:
- Updated gitignore with generated directories (models, open-jury-eval-data, results)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,8 +8,3 @@ wheels/
 
 # Virtual environments
 .venv
-
-# Evaluation data and results
-models/
-openjury-eval-data/
-results/


.gitignore
```diff
@@ -8,3 +8,8 @@ wheels/
 
 # Virtual environments
 .venv
+
+# Evaluation data and results
+models/
```
geoalgo: If I am not mistaken files should be written at $OPENJURY_EVAL_DATA in particular openjury-eval-data/ which defaults to ~/openjury-eval-data/ so it should not be in the gitignore?

Regarding `models/` and `results/` I would also remove them given that they are only related to the llamacpp example.

geoalgo (commented):
Thanks LGTM, only have a small comment about the gitignore.

PR #21, opened 2026-03-02
feat: add support for the original mt-bench

ErlisLushtaku:
## Summary

This PR implements full multi-turn evaluation support for the original MT-Bench benchmark . It aligns the turn 2 prompt structure with the methodology recommended in the original paper ("[Judging LLM-as-a-Judge](https://arxiv.org/pdf/2306.05685)") to prevent judge confusion and ensure high-quality evaluations.

## Key Changes

* **Dataset & Loader**:
  * Implemented `mt_bench.py` to handle downloading questions and reference answers from the LMSYS HuggingFace space.

* **Prompt Alignment**:
  * Created dedicated multi-turn templates (`prompt-multi-turn.txt`) that present two separate, full conversation histories (`User -> A1 -> User -> A2` and `User -> B1 -> User -> B2`).
  * This prevents the "attribution error" where the judge misidentifies model responses between turns (see paper section 3.5).

* **Pipeline Logic**:
  * Added `_run_mt_bench()` to perform separate evaluation calls for Turn 1 and Turn 2.
  * Integrated **Reference-guided grading** for technical categories (Math, Reasoning, Coding) by embedding ground-truth answers directly into the judge context.

* **Model Wrappers**:
  * Refactored `LlamaCpp` into `ChatLlamaCppModel` to support GGUF chat templates and proper KV cache management between multi-turn calls.
  * Extracted `BaseLocalModel` to share logic between LlamaCpp and VLLM wrappers.

* **Features & Metrics**:
  * Added `--mt_bench_turns` flag to support single-turn, multi-turn, or both.
  * Added `--mt_bench_compatibility` argument, which when set to `fastchat` reproduces the implementation of the paper from [FastChat/MT-Bench](https://github.com/lm-sys/FastChat/tree/main/fastchat/llm_judge) with [[A]]/[[B]]/[[C]] verdict parsing, conservative position-bias handling (a model wins only if both orderings agree, otherwise it is a tie), judge temperature=0, and MT-Bench category temperatures. If set to `openjury` (default) then it uses the OpenJury evaluation with softmax, uses averaging for position bias, doesn't set judge temperature, doesn't use per-category temperature configs.
  * Updated results output to include per-category and per-turn win rate breakdowns.

## Testing
* Added comprehensive unit and integration tests in `tests/test_generate_and_evaluate.py`.

---

> [!NOTE]
> **Medium Risk**
> Medium risk due to substantial new evaluation pipeline logic (multi-turn generation/judging, swap-mode handling) and refactoring of model wrappers/CLI parsing that could affect existing evaluation runs.
> 
> **Overview**
> Adds full `mt-bench` support end-to-end: downloads MT-Bench questions (and GPT-4 references when available), generates two-turn completions, and evaluates turn 1/turn 2 with per-category/per-turn result breakdowns.
> 
> Introduces a dedicated MT-Bench pipeline (`openjury/mt_bench`) with two judging modes: default OpenJury `score_A/score_B` parsing and an optional FastChat-compatible mode (`--mt_bench_compatibility fastchat`) using `[[A]]/[[B]]/[[C]]` verdict parsing, category temperature presets, and conservative swap-mode bias handling.
> 
> Refactors common runtime pieces by extracting CLI parsing into `openjury/config.py`, shared judging/stats into `openjury/eval_runtime.py`, adds shared `truncate`/`safe_text` helpers, and extends local provider wrappers (notably adding a `LlamaCpp` chat wrapper and temperature control) to support the new multi-turn workflows; updates docs and tests accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0094eea47dc1fcc0fddbfaa21a1520693a385db9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


geoalgo (commented):
I just did a first pass and had a few comments, I will finish the code review this afternoon.

geoalgo (commented):
Thanks look clean overall, I just have a question regarding having the feature in a separate endpoint.

We can leave it like this for now but perhaps we would want to have separate endpoint for multiturn generation and evaluation as it may be cleaner as the code is quite different? (I mean to have a `generate_and_evaluate_multiturn.py` as the entrypoint may blowup otherwise)

Also I assume the complexity of multiturn is going to increase a bit when we merge the other PR. Let me know what you think.

openjury/mt_bench/pipeline.py
```diff
@@ -0,0 +1,477 @@
+"""MT-Bench evaluation pipeline.
```
geoalgo: given that this file does not only contains pipeline but a bunch of utils, mt_bench_utils.py would probably be a better name.

PR #31, opened 2026-04-02
Unify CLI configuration & deduplicate `truncate`

kargibora:
## Problem

The codebase had two kinds of duplication that made maintenance error-prone:

1. **Two independent `truncate` implementations** — one as a top-level function in `generate.py` and another nested inside `annotate_battles()` in `evaluate.py`. The two behaved differently: the nested version handled non-string inputs gracefully while the top-level one did not. A bug fix in one would never reach the other.

2. **12 identical CLI flags copy-pasted across entrypoints** — `generate_and_evaluate.py` and `estimate_elo_ratings.py` each defined the same `--engine`, `--judge`, `--swap_mode`, `--max_new_tokens`, etc. with their own argparse blocks and dataclass fields. Any default change or new shared flag required editing both files in lock-step, with no compiler or test to catch drift.

Additionally, `swap_mode` accepted arbitrary strings with no validation — a typo like `"boht"` would silently propagate through the pipeline and produce unexpected results.

## Solution

- **`judgearena/cli_common.py`** introduces `BaseCliArgs` (a dataclass with the 12 shared fields), `add_common_arguments()` (a single argparse registration point), and `parse_engine_kwargs()` (safe JSON parsing). `BaseCliArgs.__post_init__` validates `swap_mode ∈ {"fixed", "both"}` at construction time.

- **`judgearena/utils.py`** now owns the single canonical `truncate(s, max_len)` using the safer non-string guard from `evaluate.py`.

- Both entrypoints (`generate_and_evaluate.py`, `estimate_elo_ratings.py`) inherit from `BaseCliArgs` and declare only their unique fields, cutting ~200 lines of duplicated argument definitions.

## Files changed

| File | What changed |
|---|---|
| `judgearena/cli_common.py` | **New** — `BaseCliArgs`, `add_common_arguments()`, `parse_engine_kwargs()` |
| `judgearena/utils.py` | Added canonical `truncate()` |
| `judgearena/generate.py` | Removed local `truncate()`; imports from `utils` |
| `judgearena/evaluate.py` | Removed nested `truncate()`; imports from `utils` |
| `judgearena/generate_and_evaluate.py` | `CliArgs` → inherits `BaseCliArgs` (4 unique fields remain) |
| `judgearena/estimate_elo_ratings.py` | `CliEloArgs` → inherits `BaseCliArgs` (7 unique fields remain); removed unused `json`/`field` imports |

## Testing

```
$ uv run pytest
27 passed, 2 warnings in 60.85s
```

The two warnings are pre-existing deprecations in `transformers` and `langchain` — unrelated to this change.


judgearena/estimate_elo_ratings.py
```diff
@@ -1,46 +1,30 @@
...
-class CliEloArgs:
-    arena: str
-    model: str
-    judge_model: str
-    n_instructions: int | None = None
+class CliEloArgs(BaseCliArgs):
+    """CLI arguments for the ELO rating estimation entrypoint."""
+
+    arena: str = ""
+    model: str = ""
```
geoalgo:
```suggestion
    arena: str | None = None
    model: str | None = None
```

judgearena/estimate_elo_ratings.py
```diff
@@ -1,46 +1,30 @@
...
 from judgearena.utils import cache_function_dataframe, compute_pref_summary, make_model
 
 
 @dataclass
-class CliEloArgs:
-    arena: str
-    model: str
-    judge_model: str
-    n_instructions: int | None = None
+class CliEloArgs(BaseCliArgs):
```
geoalgo, line 18: Note: We pay the price of inheriting a dataclass by requiring all members to have default bellow even if they do not make sense (we cannot have `arena` not initialized).
This is OK but we may want to avoid dataclass in the future if it becomes too messy.
Could you leave a comment indicating this?

judgearena/estimate_elo_ratings.py
```diff
@@ -137,79 +77,19 @@ def parse_args(cls):
...
-        parser.add_argument(
-            "--engine_kwargs",
-            type=str,
-            required=False,
-            default="{}",
-            help=(
-                "JSON dict of engine-specific kwargs forwarded to the underlying engine. "
-                'Example for vLLM: \'{"tensor_parallel_size": 2, "gpu_memory_utilization": 0.9}\'.'
-            ),
-        )
```
geoalgo, line 196: nice :)

judgearena/generate_and_evaluate.py
```diff
@@ -64,30 +65,13 @@ def try_load_dataset_completions(
...
-    judge_model: str
-
-    n_instructions: int | None = None
-    provide_explanation: bool = False
-    swap_mode: str = "fixed"
-    ignore_cache: bool = False
+class CliArgs(BaseCliArgs):
+    """CLI arguments for the generate-and-evaluate entrypoint."""
+
+    dataset: str = ""
```
geoalgo: same here, better have `str | None = None`

geoalgo (commented):
This is a good PR thanks @kargibora ! Just have small comments. 
Also @ErlisLushtaku can you let us know if this PR will impact your workflow?
If not, happy to merge after my small comments are addressed.

judgearena/cli_common.py
```diff
@@ -0,0 +1,171 @@
...
+    )
+    parser.add_argument(
+        "--swap_mode",
+        type=str,
+        choices=["fixed", "both"],
+        default="fixed",
+        help=(
+            "Model comparison order mode. 'fixed': always use model order A-B. "
+            "'both': correct for model order bias by evaluating each instruction "
+            "twice, once as A-B and once as B-A, and average. This helps account "
```
ErlisLushtaku: nit: We do not actually average (we concatenate), could you please update the description to not mention "averaging" same as [here](https://github.com/OpenEuroLLM/JudgeArena/pull/31/changes#diff-86d01112d69e8c8f8478bcfb4cf8c1c4fee859477facb66580f15c0370d7605cL135)

judgearena/cli_common.py
ErlisLushtaku, line 1: I had added a config file as part of the mt-bench PR: https://github.com/OpenEuroLLM/JudgeArena/pull/21/changes#diff-ee48e912ce7c7303506e9f981cc6fb9ade571a043a6acbc6fa77de513b9248fd

Could you please merge main and remove that so we don't duplicate it

judgearena/utils.py
```diff
@@ -69,6 +69,19 @@ def compute_pref_summary(prefs: pd.Series) -> dict[str, float | int]:
...
+    Non-string inputs (e.g. ``None`` or ``float('nan')``) are coerced to the
+    empty string so that callers don't have to guard against missing data.
+    """
+    if not isinstance(s, str):
+        return ""
+    if max_len is not None:
+        return s[:max_len]
+    return s
+
+
```
ErlisLushtaku, line 111: Also we added an additional function safe_text after this [here](https://github.com/OpenEuroLLM/JudgeArena/pull/21/changes#diff-ae57762b56272f6dbe91479eb39f3e2a9c6279eb41495fe55d19e1330321a633R52-R68), make sure we don't lose that after merging main, as it would break mt-bench

ErlisLushtaku (commented):
LGTM, just a couple of comments to make sure we don't break or duplicate anything after getting the mt-bench changes from main

judgearena/utils.py
kargibora: We can merge this branch after the mt-bench or vice-versa and simply insert `safe_text`

kargibora: I have inserted `safe_text` thanks for the head up

ErlisLushtaku (approved):

ErlisLushtaku:
@kargibora can we merge this?

kargibora:
> @kargibora can we merge this?

It seems we have a conflict now. I will fix it and merge it

PR #32, opened 2026-04-06
Changes related to running benchmark experiments for the paper: Support Qwen3.5 and thinking models, Skywork, truncation tracking, benchmark changes etc

ErlisLushtaku:
- Updated dependencies to support Qwen3.5
- Added thinking token budget to prevent Qwen from using all budget without outputting verdict.
- ...


pyproject.toml
```diff
@@ -81,5 +81,6 @@ quote-style = "double"
 indent-style = "space"
 
 [project.optional-dependencies]
-vllm = ["vllm==0.10.2", "transformers>=4.55.2,<5.0.0"]
+# vLLM on PyPI pins transformers<5; optional extra matches that so `uv lock` can resolve.
+vllm = ["vllm>=0.17.0,<1.0.0", "transformers>=4.56.0,<5.0.0"]
```
kargibora: `vllm>=0.17.0,<1.0.0` is a very wide range. A few concerns:

- Was this tested with a **prebuilt wheel** or built from source? Building vLLM from source on cluster nodes often fails due to CUDA kernel compilation issues.
- Is the `StructuredOutputsParams` import path (`vllm.sampling_params`) stable across this entire range? It may have been introduced in 0.17 and could move. For example `StructuredOutputParams` was a bit different when `vllm==0.11.0`. Thus I think it makes more sense to create more stable versioning

judgearena/evaluate.py
```diff
@@ -51,6 +52,18 @@ def get_regexp_match(self, s: str, regex: str, group_index: int = 1):
             return float(m.group(group_index).strip(" "))
 
 
+_PAIR_SCORE_MIN = 0
+_PAIR_SCORE_MAX = 10
+
+
+def build_pair_score_output_choices() -> list[str]:
```
kargibora: The cartesian product approach works for a single A-vs-B pair (11×11 = 121 choices), but won't scale to multi-criteria evaluation — with *N* dimensions it becomes 11^(2N) choices, which is unusable.

May be we can consider switching to a JSON schema constraint instead of `choice`, e.g. `{"score_A": int, "score_B": int}` per criterion. VLLM's `StructuredOutputsParams` already supports `json_schema` alongside `choice`, so this would be a drop-in change.

judgearena/evaluate.py
```diff
@@ -302,27 +315,30 @@ def annotate_battles(
...
+                "completion_B": truncated_completion_B,
             }
-            for user_prompt, completion_A, completion_B in zip(
-                instructions, completions_A, completions_B, strict=True
-            )
-        ]
-    )
+        )
+    if truncated_completion_count:
+        print(
```
kargibora: Flagging for a follow-up PR: the codebase mixes `print()` for warnings, progress, and debug info, making it hard to filter by severity or redirect output. We should migrate to Python's `logging` module (or at minimum a thin wrapper like `logger = logging.getLogger(__name__)`). What do you think @geoalgo

pyproject.toml
ErlisLushtaku: Good point. I tightened the range. 0.18.1 was working. I think the `StructuredOutputParams` is stable accross the new range.

judgearena/evaluate.py
ErlisLushtaku: Agreed, updated

pyproject.toml
ErlisLushtaku: Changed it to `v0.19+` so that we can use the thinking token limit parameter, and also they have some fixes for Qwen3.5

judgearena/utils.py
```diff
@@ -286,11 +494,23 @@ def __init__(
...
                 self._use_generate = True
+                if disable_thinking:
+                    warnings.warn(
+                        f"Model '{model}' has no chat template, so disable_thinking "
+                        "cannot be applied when falling back to llm.generate().",
+                        stacklevel=2,
+                    )
             else:
-                self.chat_template = None  # let vLLM use the tokenizer's own
+                self.chat_template = None
```
ErlisLushtaku:
```suggestion
                self.chat_template = None  # let vLLM use the tokenizer's own
```

geoalgo:
Closed as it was splitted in independent PRs.

PR #34, opened 2026-04-07
add LMArena-55k

geoalgo:


judgearena/arenas_utils.py
ErlisLushtaku: Should "LMArena-55k" be added here too?

ErlisLushtaku (approved):
Looks good, just a small question

judgearena/arenas_utils.py
geoalgo: good catch thanks, sent an edit.

PR #35, opened 2026-04-08
small features: add option to save cache in parquet, save judge input…

geoalgo:
…, improve error handling of openrouter, remove compute_cohen_kappa

Reasons:
- saving in parquet is better in term of storage
- saving judge input allow to have all the context used to make the judge call and to be able to estimate the cost 
- the handling of errors in openrouter becomes a bit less whacky
- cohen_kappa is available in scikit-learn, we should use this code instead (which produces the same values)

### DIFF
--- a/judgearena/evaluate.py
+++ b/judgearena/evaluate.py
@@ -252,10 +252,11 @@ def evaluate_completions(
 
 @dataclass
 class JudgeAnnotation:
-    judge_completion: str
-    instruction: str
-    completion_A: str
-    completion_B: str
+    instruction: str  # instruction from the user
+    completion_A: str  # completion of the first model
+    completion_B: str  # completion of the second model
+    judge_completion: str  # output of the judge
+    judge_input: str | None = None  # input that was passed to the judge
 
 
 def annotate_battles(
@@ -338,11 +339,17 @@ def annotate_battles(
     )
 
     annotations = []
-    for judge_completion, instruction, completion_A, completion_B in zip(
-        judge_completions, instructions, completions_A, completions_B, strict=True
+    for judge_input, judge_completion, instruction, completion_A, completion_B in zip(
+        inputs,
+        judge_completions,
+        instructions,
+        completions_A,
+        completions_B,
+        strict=True,
     ):
         annotations.append(
             JudgeAnnotation(
+                judge_input=judge_input,
                 judge_completion=judge_completion,
                 instruction=instruction,
                 completion_A=completion_A,
--- a/judgearena/utils.py
+++ b/judgearena/utils.py
@@ -91,6 +91,28 @@ def compute_pref_summary(prefs: pd.Series) -> dict[str, float | int]:
     }
 
 
+def _is_retryable_error(e: Exception) -> bool:
+    """Return True if the exception is a transient server error that should be retried.
+
+    Handles two formats:
+    - String representation contains the HTTP code (most providers)
+    - ValueError raised by langchain-openai with a dict arg: {'message': ..., 'code': 429}
+    """
+    # langchain-openai raises ValueError(response_dict.get("error")) where the
+    # error value is a dict like {'message': '...', 'code': 408}
+    _RETRYABLE_CODES = {408, 429, 502, 503, 504}
+    if isinstance(e, ValueError) and e.args:
+        arg = e.args[0]
+        if isinstance(arg, dict) and arg.get("code") in _RETRYABLE_CODES:
+            return True
+
+    error_str = str(e)
+    return (
+        any(str(code) in error_str for code in _RETRYABLE_CODES)
+        or "rate" in error_str.lower()
+    )
+
+
 def do_inference(chat_model, inputs, use_tqdm: bool = False):
     # Retries on rate-limit/server errors with exponential backoff.
     # Async path retries individual calls; batch path splits into 4^attempt chunks on failure.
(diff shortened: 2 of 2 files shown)


judgearena/evaluate.py
```diff
@@ -252,10 +252,11 @@ def get_output(df_outputs: pd.DataFrame, dataset: str, method: str):
...
 class JudgeAnnotation:
-    judge_completion: str
-    instruction: str
-    completion_A: str
-    completion_B: str
+    instruction: str  # instruction from the user
+    completion_A: str  # completion of the first model
+    completion_B: str  # completion of the second model
+    judge_completion: str  # output of the judge
+    judge_input: str | None = None  # input that was passed to the judge
```
ErlisLushtaku, line 259: Should this be added to the `estimate_elo_ratings.py` workflow as well?

judgearena/utils.py
```diff
@@ -481,80 +495,60 @@ def cache_function_dataframe(
...
+                    if isinstance(x, np.ndarray):
+                        return [_to_python(i) for i in x]
+                    if isinstance(x, dict):
+                        return {k: _to_python(v) for k, v in x.items()}
+                    if isinstance(x, list):
+                        return [_to_python(i) for i in x]
+                    return x
+
+                for col in df.select_dtypes(include="object").columns:
+                    df[col] = df[col].apply(_to_python).astype(str)
```
ErlisLushtaku, line 546: I think we should be careful here if the dataframe can contain missing values. Calling `.astype(str)` on missing values (None or np.nan) converts them into strings "None" and "nan". When the parquet file is read back, they would be processed as strings instead of missing values.

judgearena/utils.py
```diff
@@ -91,6 +91,28 @@ def compute_pref_summary(prefs: pd.Series) -> dict[str, float | int]:
...
+def _is_retryable_error(e: Exception) -> bool:
+    """Return True if the exception is a transient server error that should be retried.
+
+    Handles two formats:
+    - String representation contains the HTTP code (most providers)
+    - ValueError raised by langchain-openai with a dict arg: {'message': ..., 'code': 429}
+    """
+    # langchain-openai raises ValueError(response_dict.get("error")) where the
+    # error value is a dict like {'message': '...', 'code': 408}
+    _RETRYABLE_CODES = {408, 429, 502, 503, 504}
```
ErlisLushtaku, line 103: We are missing code `500` here compared to the [previous behavior](https://github.com/OpenEuroLLM/JudgeArena/pull/35/changes#diff-ae57762b56272f6dbe91479eb39f3e2a9c6279eb41495fe55d19e1330321a633L149). Do we want to remove that code?

judgearena/evaluate.py
geoalgo: It uses `judge_and_parse_prefs` from this file so it is updated as well if I am not mistaken.

judgearena/utils.py
geoalgo: yes I agree but I dont see another way to serialize to parquet. I agree that this conversion is loosing the missingness information but I think all downstream code should probably exclude empty strings too when computing annotations.

judgearena/utils.py
geoalgo: It was a typo, I added only the error which seems retryable from here:
https://openrouter.ai/docs/api/reference/errors-and-debugging

judgearena/evaluate.py
ErlisLushtaku: Yes, but we are dropping it [here](https://github.com/OpenEuroLLM/JudgeArena/blob/ee049248745313bfe14f45cd54cc44e6f254afa7/judgearena/estimate_elo_ratings.py#L484) because we are constructing the Dataframe manually

judgearena/utils.py
ErlisLushtaku: I searched a bit and I think we have two safe solutions:

Option 1: Parquet + JSON Sidecar
We selectively stringify only the complex objects (like dicts/lists), leave None/NaN untouched by filtering something like this:
```
df_cache[col] = df_cache[col].apply(
    lambda x: x if x is None or (isinstance(x, float) and np.isnan(x)) else repr(_to_python(x))
)
```
and save a small `.meta.json` tracking which columns were altered. Then when reading we again filter to call `ast.literal_eval` only on cells that don't contain these missing values and were serialized.

Option 2: Compressed Pickle (.pkl.gz)
We could switch to `df.to_pickle(cache_file, compression="gzip")`. It would be simpler, but probably less storage-efficient.

Otherwise we could leave it like this and note it for later, if you think this is not a big issue at the moment.

geoalgo: Option 1 seems like an overkill given that the option is for util which is deactivated by default.
Option 2 is a strong no, as we would be replacing the issue with a security issue (also we wont be getting the benefit of parquet of column storage compression).

I would be keen to merge as is, possibly adding a comment. 
Alternatively, I am also fine to drop the feature and keep it in my branch.

judgearena/evaluate.py
geoalgo: good catch, I will change the elo-rating to also store input.

judgearena/utils.py
ErlisLushtaku: Then we could merge it as is so we can use it too 👍

ErlisLushtaku (approved):

PR #38, opened 2026-04-24
Fix bug with Vllm specific args, unify CLI and rename arguments

ErlisLushtaku:


geoalgo (commented):
LGTM to merge, we need just need to fix ruff.

geoalgo (approved):

PR #39, opened 2026-04-29
Pin dataset revisions for reproducibility

alexrs-cohere:
## Summary

- Adds `judgearena/dataset_revisions.py` as the single source of truth for every HuggingFace dataset/space and raw URL the library downloads. `HF_DATASET_REVISIONS` maps repo IDs to commit SHAs and `RAW_URL_REVISIONS` does the same for raw GitHub URLs (currently only the FastChat MT-Bench reference answers).
- Forwards `revision=` through every `snapshot_download` / `load_dataset` / `urlretrieve` call:
  - `judgearena/arenas_utils.py` — LMArena 55k / 100k / 140k, plus the existing ComparIA pin is moved into the new module so the default revision is now derived from `hf_revision("ministere-culture/comparia-votes")`.
  - `judgearena/instruction_dataset/arena_hard.py` — `lmarena-ai/arena-hard-auto`.
  - `judgearena/instruction_dataset/m_arenahard.py` — `CohereLabs/m-ArenaHard`.
  - `judgearena/instruction_dataset/mt_bench.py` — `lmsys/mt-bench` space and the FastChat raw-URL via a new `_fastchat_reference_url()` helper.
  - `judgearena/utils.py` — `geoalgo/llmjudge` (AlpacaEval) and `geoalgo/multilingual-contexts-to-be-completed`.
- The `judgearena.utils` import of `judgearena.instruction_dataset.arena_hard` moves to the bottom of the module to break the import cycle introduced by `m_arenahard.py` now importing `data_root`.

The pinned table currently keeps the existing ComparIA pin and leaves the remaining repos as ``None`` until we choose SHAs, but every download site is already plumbed so bumping a pin is a one-line change in ``dataset_revisions.py``. When a value is ``None`` the call passes ``revision=None``, i.e. unchanged behaviour.

A follow-up PR will record the resolved revisions under ``metadata["dataset_revisions"]`` so each run captures which version of the data it actually saw.

## Why

Today only ComparIA is pinned (via a magic string in ``arenas_utils.py``). The other six download sites silently track upstream ``main``, so a run that succeeds today can produce different bytes — and therefore different results — after any upstream rewrite, contradicting the paper's reproducibility claims. This PR is the smallest piece of plumbing that lets us pin once we choose SHAs.

## Test plan

- [x] ``uv run pytest -q`` passes (74 tests on this branch in isolation).
- [ ] CI runs the full suite.
- [ ] Sanity-check ``download_all()`` end-to-end on a clean ``~/judgearena-data`` to confirm ``revision=None`` keeps existing behaviour untouched.

### DIFF
--- a/judgearena/arenas_utils.py
+++ b/judgearena/arenas_utils.py
@@ -5,6 +5,7 @@ import pandas as pd
 from fast_langdetect import detect_language
 from huggingface_hub import snapshot_download
 
+from judgearena.dataset_revisions import hf_revision
 from judgearena.log import get_logger
 
 logger = get_logger(__name__)
@@ -30,11 +31,13 @@ def _load_arena_dataframe(
 ) -> pd.DataFrame:
     assert arena in KNOWN_ARENAS
     if arena == "LMArena-55k":
+        repo_id = "lmarena-ai/arena-human-preference-55k"
         path = snapshot_download(
-            repo_id="lmarena-ai/arena-human-preference-55k",
+            repo_id=repo_id,
             repo_type="dataset",
             allow_patterns="*.csv",
             force_download=False,
+            revision=hf_revision(repo_id),
         )
         df = pd.read_csv(Path(path) / "train.csv")
 
@@ -70,11 +73,13 @@ def _load_arena_dataframe(
 
     elif "LMArena" in arena:
         size = arena.split("-")[1]  # "100k" or "140k"
+        repo_id = f"lmarena-ai/arena-human-preference-{size}"
         path = snapshot_download(
-            repo_id=f"lmarena-ai/arena-human-preference-{size}",
+            repo_id=repo_id,
             repo_type="dataset",
             allow_patterns="*parquet",
             force_download=False,
+            revision=hf_revision(repo_id),
         )
         parquet_files = sorted((Path(path) / "data").glob("*.parquet"))
         df = pd.concat([pd.read_parquet(f) for f in parquet_files], ignore_index=True)
@@ -171,9 +176,12 @@ def _load_arena_dataframe(
     return df
 
 
+_DEFAULT_COMPARIA_REVISION = hf_revision("ministere-culture/comparia-votes")
+
+
 def load_arena_dataframe(
     arena: str | None,
-    comparia_revision: str = "7a40bce496c1f2aa3be4001da85a49cb4743042b",
+    comparia_revision: str | None = _DEFAULT_COMPARIA_REVISION,
 ) -> pd.DataFrame:
     """Load battles from one or all arenas.
 
--- a/judgearena/dataset_revisions.py
+++ b/judgearena/dataset_revisions.py
@@ -1,0 +1,64 @@
+"""Pinned upstream revisions for every dataset/space JudgeArena downloads.
+
+Pinning lets the run metadata answer "exactly which version of the data did
+this run see?".  When upstream rewrites a dataset (e.g. ComparIA periodically
+republishes), an unpinned ``snapshot_download`` will silently start returning
+different bytes; pinned revisions force callers to opt into upgrades.
+
+To bump a revision, paste the new commit SHA from the dataset's HuggingFace
+revision page (or the GitHub commit page for the FastChat raw URL).
+"""
+
+from __future__ import annotations
+
+# HuggingFace dataset / space revisions.  Keys are HuggingFace ``repo_id``
+# strings; values are commit SHAs.  ``None`` is allowed for repos where we
+# do not yet have a stable pin and is recorded as such in the metadata so
+# the gap is visible.
+HF_DATASET_REVISIONS: dict[str, str | None] = {
+    # LMArena human-preference battles
+    "lmarena-ai/arena-human-preference-100k": "72e85b3ddc9c81bf7b659d6b03d4126dfd8fb34a",
+    "lmarena-ai/arena-human-preference-140k": "6322995ab34d7c2693e3f47dd13fa5caa0789a74",
+    "lmarena-ai/arena-human-preference-55k": "18c298340948c0e7f7727399fd459cca6ce0ca6f",
+    # ComparIA (already pinned via the legacy comparia_revision argument).
+    "ministere-culture/comparia-votes": "7a40bce496c1f2aa3be4001da85a49cb4743042b",
+    # m-ArenaHard (Cohere release)
+    "CohereLabs/m-ArenaHard": "ab393a96cd0b134a1acfa96e080af31e5e73a393",
+    "CohereLabs/m-ArenaHard-v2.0": "24c65eff42cec85e30dd5db99d1a702c7ebaa8ab",
+    # AlpacaEval instructions / model_outputs (geoalgo redistribution; the
+    # repo now redirects to ``judge-arena/judge-arena-dataset`` upstream, but
+    # ``snapshot_download`` follows the redirect transparently).
+    "geoalgo/llmjudge": "004c4a992956eeefffd36b63ade470f32fd0a582",
+    # MT-Bench questions (LMSYS Space).
+    "lmsys/mt-bench": "a4b674ca573c24143824ac7f60d9173e7081e37d",
+    # Multilingual fluency contexts.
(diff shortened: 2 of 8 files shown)


judgearena/arenas_utils.py
```diff
@@ -171,9 +176,15 @@ def get_winner(
     return df
 
 
+_DEFAULT_COMPARIA_REVISION = (
+    hf_revision("ministere-culture/comparia-votes")
```
geoalgo: Why do we need the `or` given that hf_revision("ministere-culture/comparia-votes") is defined?

judgearena/dataset_revisions.py
```diff
@@ -0,0 +1,61 @@
...
+    # m-ArenaHard (Cohere release)
+    "CohereLabs/m-ArenaHard": None,
+    # AlpacaEval instructions / model_outputs (geoalgo redistribution).
+    "geoalgo/llmjudge": None,
+    # MT-Bench questions (LMSYS Space).
+    "lmsys/mt-bench": None,
+    # Multilingual fluency contexts.
+    "geoalgo/multilingual-contexts-to-be-completed": None,
+    # Arena-Hard official source (used via datasets.load_dataset).
+    "lmarena-ai/arena-hard-auto": None,
```
geoalgo: Can we tied all of those to the current commit as well? 
I agree its probably better than to have those updated silently.

judgearena/utils.py
```diff
@@ -40,12 +40,14 @@ def set_langchain_cache():
 def download_hf(name: str, local_path: Path):
     local_path.mkdir(exist_ok=True, parents=True)
     # downloads the model from huggingface into `local_path` folder
+    repo_id = "geoalgo/llmjudge"
```
geoalgo: Can we avoid changes which are not related to what the PR want to introduce?

judgearena/utils.py
```diff
@@ -482,12 +484,14 @@ def download_all():
         else:
             download_hf(name=dataset, local_path=local_path_tables)
 
+    contexts_repo = "geoalgo/multilingual-contexts-to-be-completed"
```
geoalgo, line 724: Same here.

judgearena/utils.py
```diff
@@ -582,5 +586,15 @@ def _to_python(x):
                 return pd.read_csv(cache_file)
 
 
+# Imported at the bottom so that ``data_root``, ``download_hf`` and ``read_df``
```
geoalgo: Can we rather move things in an util? It is better to import at the top of files.

geoalgo (commented):
LGTM overall, I left some comments.

ErlisLushtaku:
@alexrs-cohere Thanks for the PR Alex. There is a bit of an issue because I have been working for some time on this [big PR](https://github.com/OpenEuroLLM/JudgeArena/pull/32#top), and it would be better if you can rebase your PRs there until we merge it, since otherwise I would have to redo most things because a lot has changed.

geoalgo:
This PR should be rather independent from your changes @ErlisLushtaku right?

In any cases, I do not think we should rebase on the other PR as its 7K lines added, we would want to trim it down before.

judgearena/utils.py
alexrs-cohere: This is done to avoid repetitions later in `snapshot_download` as we need to specify `repo_id=...` and `revision=hf_revision(...),`

judgearena/utils.py
alexrs-cohere: https://github.com/OpenEuroLLM/JudgeArena/pull/39/changes#r3165448983

judgearena/dataset_revisions.py
alexrs-cohere: What SHAs were used for ablations?

judgearena/arenas_utils.py
alexrs-cohere: Fixed!

judgearena/dataset_revisions.py
alexrs-cohere: Fixed! Got the latest SHAs for each repo

ErlisLushtaku:
> This PR should be rather independent from your changes @ErlisLushtaku right?
> 
> In any cases, I do not think we should rebase on the other PR as its 7K lines added, we would want to trim it down before.

The feature is independent in intent, but it no longer applies cleanly on top of my branch. There would be some conflicts. However that is fine if you want to merge this first, until I clean that branch up.

judgearena/dataset_revisions.py
geoalgo: > What SHAs were used for ablations?

We tagged comparia as this dataset was changing all the time (which is great). The other have not received any commit as they were static dumps as far as I know.

uv.lock
geoalgo, line 1: I dont think we should commit this given that this is a library published in pypi, we have a pyproject.toml to specify our dependencies.

geoalgo:
LGTM, can you just remove the `uv.lock`?

uv.lock
alexrs-cohere: The Astral docs specifically recommend to track the `uv.lock` file -- https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile

> This file should be checked into version control, allowing for consistent and reproducible installations across machines.
>
> A lockfile ensures that developers working on the project are using a consistent set of package versions. Additionally, it ensures when deploying the project as an application that the exact set of used package versions is known.

This file should not interfere with installations from pypi. When we run `pip install judgearena`, pip ignores the lock file entirely. pip resolves transitively from pyproject.toml's dependencies = [...] against the user's environment. The lock has zero effect on what your downstream users get.

Adding the lockfile helps during development so we all use the same dependencies, and CI runs are consistent.

If you still disagree, I'm happy to remove it!

geoalgo: I am not sure reading about this, for instance see https://github.com/astral-sh/uv/issues/10730 lets not do that in this PR as it is not its intent, we can revisit this discussion in an issue.

alexrs-cohere: Done!

alexrs-cohere: https://github.com/astral-sh/uv/issues/10730#issuecomment-2601604397 however confirms that the `uv.lock` does not have impact on libraries.

geoalgo:
Need to rebase against mainline, @kargibora will take care of this.

alexrs-cohere:
@geoalgo @ErlisLushtaku I will take a look at the comments on this PR and the others in the next few days!

geoalgo:
ah cool to know, let us know if you need help to rebase your branch.

judgearena/utils.py
alexrs-cohere: Keeping the local variable here intentionally: `contexts_repo` is used both as the `repo_id` and as the key into `hf_revision(...)`, so the two stay coupled if this repo id changes later. This mirrors the other pinned `snapshot_download` sites in this PR.

judgearena/dataset_revisions.py
```diff
@@ -0,0 +1,64 @@
...
+HF_DATASET_REVISIONS: dict[str, str | None] = {
+    # LMArena human-preference battles
+    "lmarena-ai/arena-human-preference-100k": "72e85b3ddc9c81bf7b659d6b03d4126dfd8fb34a",
+    "lmarena-ai/arena-human-preference-140k": "6322995ab34d7c2693e3f47dd13fa5caa0789a74",
+    "lmarena-ai/arena-human-preference-55k": "18c298340948c0e7f7727399fd459cca6ce0ca6f",
+    # ComparIA (already pinned via the legacy comparia_revision argument).
+    "ministere-culture/comparia-votes": "7a40bce496c1f2aa3be4001da85a49cb4743042b",
+    # m-ArenaHard (Cohere release)
+    "CohereLabs/m-ArenaHard": "ab393a96cd0b134a1acfa96e080af31e5e73a393",
+    "CohereLabs/m-ArenaHard-v2.0": "24c65eff42cec85e30dd5db99d1a702c7ebaa8ab",
```
alexrs-cohere, line 27: We also have m-ArenaHard-v2.1 now! Can be added as a follow-up PR.  https://huggingface.co/datasets/CohereLabs/m-ArenaHard-v2.1

ErlisLushtaku:
@geoalgo can we merge this?

geoalgo (approved):

Issue #36, opened 2026-04-08
Caching judge completion


geoalgo:
Currently, we use `set_langchain_cache()` to cache completions which does not work with vLLM for some reason and offer little control. 

I would propose to support a caching mechanism for judge completions, e.g. when calling `annotate_battles` 
https://github.com/OpenEuroLLM/JudgeArena/blob/main/judgearena/evaluate.py#L261 which caches judge-annotations in files like `{judge_arena_dir}/cache/db/{benchmark}/{judge}.db` (the naming would allow the user to manually delete some entries easily).

For the storage, we could use sqlite which would not require any dependency.

For the schema, we could use `"benchmark", "instruction_id", "model_a", "model_b",  "judge"` as keys and retrieve any entry from the db if we get a hit and generate otherwise. 

We could store the following which would allow to return completion from cache and perform small analysis:

```
@dataclass
class AnnotationEntry:
    benchmark: str
    instruction_id: str
    model_a: str
    model_b: str
    judge: str
    judge_input: str
    judge_completion: str
    date: str
```

Any thoughts @ErlisLushtaku @kargibora ?

ErlisLushtaku:
I think this is a good idea, moving away from set_langchain_cache() looks like the right call. 

Just some concern regarding the cache key implementation, should we add the following:
- the actual judge input, because models could generate different things on the same instruction
- parameters that can change the prompt e.g. `provide_explanation`, and the prompt version
- trucation parameters and output-affecting judge paramters

geoalgo:
My initial thought was that we are not changing much the prompt but in case it changes, I agree that having the judge_input as key would be important (alternatively, we could put some of the metadata that Bora added).

> parameters that can change the prompt e.g. provide_explanation, and the prompt version

This would not be needed if we store the judge_input.

> trucation parameters and output-affecting judge paramters

We can try to pass the metadata but this one may be hard as we basically need to serialize all our state.

ErlisLushtaku:
> My initial thought was that we are not changing much the prompt but in case it changes, I agree that having the judge_input as key would be important (alternatively, we could put some of the metadata that Bora added).
> 
> > parameters that can change the prompt e.g. provide_explanation, and the prompt version
> 
> This would not be needed if we store the judge_input.
> 
> > trucation parameters and output-affecting judge paramters
> 
> We can try to pass the metadata but this one may be hard as we basically need to serialize all our state.

[...]

PR #41, opened 2026-04-29
Per-role GenerationConfig and backend plumbing

alexrs-cohere:
## Summary

Exposes every generation knob that can affect produced text **independently for the three runtime roles** (the evaluated model (A), the baseline/opponent model (B), and the LLM judge) within the hierarchical `RunConfig` (pydantic-settings) introduced in #64, and plumbs those knobs through to each backend.

### Caching

The generated-completion cache keys in the pairwise (`generate_and_evaluate`) and MT-Bench paths now incorporate a hash of the resolved per-role generation kwargs (via `generation_cache_token`), matching the ELO path. Changing any sampling parameter (temperature, seed, top_p/k, max_tokens, chat_template, ...) now busts the cache instead of silently reusing a stale run. (Existing on-disk caches for these paths are regenerated on the next run.)

## Why

vLLM's sampling parameters were hard-coded (`temperature=0.6`, `top_p=0.95`) at the wrapper layer, hosted providers ignored the seed, and a single `max_out_tokens` was forced on both battle models even when you'd want to ablate them independently. Describing each role's generation explicitly — and keying the completion cache on it — is the foundation for hashing and replaying a run deterministically.

## Test plan

- [x] `tests/test_seed_plumbing.py`:
  - `make_model` forwards `temperature` / `seed` / `top_p` / `top_k` / arbitrary `engine_kwargs` to the backend (asserted via `DummyModel.init_kwargs`).
  - the three resolver methods skip unset fields, only override `engine_kwargs` when a dedicated field is set, keep A/B/judge kwargs independent, and inherit baseline sampling from `model.*` when unset.
  - `generation_cache_token` is order-independent and changes when any sampling param changes.
  - nested CLI flags (`--model.seed`, `--model.baseline_seed`, `--judge.temperature`) land on the correct role.
- [x] `uv run pytest -q` — full suite green (205 passed locally).
- [x] `ruff check` — clean.
- [ ] CI runs the full suite.

### DIFF
--- a/judgearena/config.py
+++ b/judgearena/config.py
@@ -24,6 +24,10 @@ _ACTIVE_CONFIG_PATH: str | None = None
 _ACTIVE_CLI_ARGS: list[str] | None = None
 
 
+def _drop_none(kwargs: dict[str, object]) -> dict[str, object]:
+    return {k: v for k, v in kwargs.items() if v is not None}
+
+
 class ModelArgs(BaseModel):
     """The model(s) under evaluation and their generation/engine settings."""
 
@@ -43,6 +47,20 @@ class ModelArgs(BaseModel):
     """Generation token budget for each evaluated-model answer (for vLLM, keep
     this <= ``max_model_len``)."""
 
+    temperature: float | None = None
+    """Sampling temperature for the evaluated model. Unset keeps the backend
+    default."""
+
+    top_p: float | None = None
+    """Nucleus-sampling probability for the evaluated model. Unset keeps the
+    backend default."""
+
+    top_k: int | None = None
+    """Top-k sampling cutoff for the evaluated model when supported."""
+
+    seed: int | None = None
+    """Backend sampling seed for the evaluated model when supported."""
+
     max_model_len: int | None = None
     """Optional total context window (prompt + generation) for the generation
     vLLM instance. Applies to vLLM models only."""
@@ -54,6 +72,107 @@ class ModelArgs(BaseModel):
     engine_kwargs: dict = Field(default_factory=dict)
     """JSON dict of engine-specific kwargs forwarded to the backend, e.g. for
     vLLM ``{"tensor_parallel_size": 2, "gpu_memory_utilization": 0.9}``."""
+
+    baseline_max_out_tokens: int | None = None
+    """Generation token budget for the baseline/opponent model. Unset inherits
+    ``model.max_out_tokens``."""
+
+    baseline_temperature: float | None = None
+    """Sampling temperature for the baseline/opponent model. Unset inherits
+    ``model.temperature``."""
+
+    baseline_top_p: float | None = None
+    """Nucleus-sampling probability for the baseline/opponent model. Unset
+    inherits ``model.top_p``."""
+
+    baseline_top_k: int | None = None
+    """Top-k sampling cutoff for the baseline/opponent model when supported.
+    Unset inherits ``model.top_k``."""
+
+    baseline_seed: int | None = None
+    """Backend sampling seed for the baseline/opponent model when supported.
+    Unset inherits ``model.seed``."""
+
+    baseline_max_model_len: int | None = None
+    """Optional total context window for the baseline/opponent vLLM instance.
+    Unset inherits ``model.max_model_len``."""
+
+    baseline_chat_template: str | None = None
+    """Jinja2 chat template for the baseline/opponent model. Unset inherits
+    ``model.chat_template``."""
+
+    baseline_engine_kwargs: dict | None = None
+    """JSON dict of engine-specific kwargs for the baseline/opponent model.
+    Unset inherits ``model.engine_kwargs``."""
+
+    def evaluated_generation_kwargs(self) -> dict[str, object]:
+        """Kwargs for model A / the evaluated model."""
+        kwargs: dict[str, object] = dict(self.engine_kwargs)
+        kwargs.update(
+            _drop_none(
+                {
+                    "max_tokens": self.max_out_tokens,
+                    "max_model_len": self.max_model_len,
+                    "chat_template": self.chat_template,
+                    "temperature": self.temperature,
+                    "top_p": self.top_p,
+                    "top_k": self.top_k,
+                    "seed": self.seed,
+                }
+            )
+        )
+        return kwargs
+
+    def baseline_generation_kwargs(self) -> dict[str, object]:
+        """Kwargs for model B / the baseline model."""
+        engine_kwargs = (
+            self.engine_kwargs
+            if self.baseline_engine_kwargs is None
+            else self.baseline_engine_kwargs
+        )
+        kwargs: dict[str, object] = dict(engine_kwargs)
+        kwargs.update(
+            _drop_none(
+                {
+                    "max_tokens": (
+                        self.baseline_max_out_tokens
+                        if self.baseline_max_out_tokens is not None
+                        else self.max_out_tokens
+                    ),
+                    "max_model_len": (
+                        self.baseline_max_model_len
+                        if self.baseline_max_model_len is not None
+                        else self.max_model_len
+                    ),
+                    "chat_template": (
+                        self.baseline_chat_template
+                        if self.baseline_chat_template is not None
+                        else self.chat_template
+                    ),
+                    "temperature": (
+                        self.baseline_temperature
+                        if self.baseline_temperature is not None
+                        else self.temperature
+                    ),
+                    "top_p": (
+                        self.baseline_top_p
+                        if self.baseline_top_p is not None
+                        else self.top_p
+                    ),
+                    "top_k": (
+                        self.baseline_top_k
+                        if self.baseline_top_k is not None
+                        else self.top_k
+                    ),
+                    "seed": (
+                        self.baseline_seed
+                        if self.baseline_seed is not None
+                        else self.seed
+                    ),
+                }
+            )
+        )
+        return kwargs
 
 
 class JudgeArgs(BaseModel):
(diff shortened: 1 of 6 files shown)


geoalgo:
Sorry for the delay, @geoalgo will rebase it and @ErlisLushtaku will then review it.

alexrs-cohere:
@geoalgo @ErlisLushtaku is this PR still relevant? I can fix all the conflicts and re-work it to get it ready if so. Let me know!

ErlisLushtaku:
> is this PR still relevant?

@alexrs-cohere yes the PR is still relevant. Thanks!

alexrs-cohere:
@ErlisLushtaku ready for review!

judgearena/config.py
```diff
@@ -55,6 +73,75 @@ class ModelArgs(BaseModel):
...
+                "max_tokens": self.max_out_tokens,
+                "max_model_len": self.max_model_len,
+                "chat_template": self.chat_template,
+                "temperature": self.temperature,
+                "top_p": self.top_p,
+                "top_k": self.top_k,
+                "seed": self.seed,
+            }
+        )
+        return _drop_none(kwargs)
```
ErlisLushtaku: This overrides any sampling key set inside engine_kwargs.the dedicated fields default to None, overwrite the engine_kwargs value, then _drop_none removes it. For example engine_kwargs={'temperature':0.5} resolves to {'max_tokens': 32768}. I think we should overlay only the set dedicated fields on top of engine_kwargs instead, e.g.:

```suggestion
    def evaluated_generation_kwargs(self) -> dict[str, object]:
        """Kwargs for model A / the evaluated model."""
        kwargs: dict[str, object] = dict(self.engine_kwargs)
       kwargs.update(_drop_none({
    "max_tokens": self.max_out_tokens, "max_model_len": self.max_model_len,
    "chat_template": self.chat_template, "temperature": self.temperature,
    "top_p": self.top_p, "top_k": self.top_k, "seed": self.seed,
}))
return kwargs
 ```

File: judgearena/utils.py [hunk 2]
```diff
@@ -233,12 +233,38 @@ def safe_parse_int(env_var: str) -> int | None:
...
+        if key in response_metadata
+    }
+    return captured or None
+
+
+def do_inference(
+    chat_model,
+    inputs,
+    use_tqdm: bool = False,
+    out_metadata: list[dict | None] | None = None,
```
ErlisLushtaku: out_metadata is never passed by any caller

judgearena/config.py
```diff
@@ -55,6 +73,75 @@ class ModelArgs(BaseModel):
...
+                "max_model_len": (
+                    self.baseline_max_model_len
+                    if self.baseline_max_model_len is not None
+                    else self.max_model_len
+                ),
+                "chat_template": self.baseline_chat_template or self.chat_template,
+                "temperature": self.baseline_temperature,
+                "top_p": self.baseline_top_p,
+                "top_k": self.baseline_top_k,
+                "seed": self.baseline_seed,
```
ErlisLushtaku: Is there a reason we are using `or` sometimes and in other cases `is not None`?

ErlisLushtaku (changes requested):
Thanks for the rebase! I have a couple of comments, also the PR description seems outdated.

alexrs-cohere:
@ErlisLushtaku Thanks for the review! I did resolve the merge conflicts but overlooked some of the changes needed after many PRs landed.

I spent some more time looking at that today, and the PR is ready for another review!

### Follow-up for fully reproducible runs

This PR gets us to the point where a run is fully described by its per-role generation settings. The natural next step is to make that description the single source of truth, so here's what I'm thinking for the next PR:

- **One canonical run descriptor** that captures everything affecting the output: the resolved generation/sampling settings, input truncation, the task and dataset revision, which instructions were used, and the judge setup (model, prompts, swap mode, etc.). We'd just need to be clear about which fields actually affect results vs. things like output folder or verbosity.
- **Content-addressed cache**: key the cache off a hash of that descriptor and use the same logic everywhere (pairwise / MT-Bench / ELO). This also fixes a remaining gap where the cache doesn't notice a dataset revision change yet.
- **`run-metadata` v2** that records the descriptor and the resolved settings (not just the raw config), plus a hash per artifact, so any run is easy to trace.
- **A `--rerun` option** that rebuilds a run straight from its metadata file and reuses the cached results.

Happy to pick these up as the next steps in the reproducibility work!

ErlisLushtaku (approved):

ErlisLushtaku:
@alexrs-cohere Thanks for the PR and the follow up suggestions.

Regarding each point:

> * **One canonical run descriptor** that captures everything affecting the output: the resolved generation/sampling settings, input truncation, the task and dataset revision, which instructions were used, and the judge setup (model, prompts, swap mode, etc.). We'd just need to be clear about which fields actually affect results vs. things like output folder or verbosity.

A good chunk of the foundation already exists RunConfig (#64) plus `repro.py` already record the resolved config, the instruction-indices hash and the judge-prompt hashes, and `dataset_revisions.py` pins the dataset revisions. There's also overlap with @kargibora's typed-results work (#66), so let's coordinate, probably easiest in a quick meeting.

> * **Content-addressed cache**: key the cache off a hash of that descriptor and use the same logic everywhere (pairwise / MT-Bench / ELO). This also fixes a remaining gap where the cache doesn't notice a dataset revision change yet.

We have this open issue #36 about the judge-completion side cache and I was looking into it. Feel free to add your thoughts. Would be more than happy to discuss this together. 

> * **`run-metadata` v2** that records the descriptor and the resolved settings (not just the raw config), plus a hash per artifact, so any run is easy to trace.

You could pick this up, just sync with @kargibora (his #63 also touches run-metadata) so we don't duplicate work.

> * **A `--rerun` option** that rebuilds a run straight from its metadata file and reuses the cached results.

This ties into point 2 and the judge-caching work (#36). `--config_path results/<run>/config.yaml` already gives a basic rerun today, so the added value here is faithful replay of the resolved settings + reusing cached judgements. We could decide to tackle this together with point 2 about judge caching or separately. Let me know what you think.

alexrs-cohere:
Thanks @ErlisLushtaku this all makes sense, and agreed on coordinating rather than landing one big PR! I've read through #64, #66, #63 and #36; here's how I'd suggest slicing it so we don't step on @kargibora's or @geoalgo's work. 

The mental model I'd propose is a two-layer descriptor, which keeps us out of the results/reporting layer entirely:

- **artifact-level descriptors** -- only the fields that change an artifact's bytes (task + resolved dataset revision + model spec + resolved per-role generation kwargs + truncation + instruction selection). Hashing these gives the cache key.
- **run-level descriptor** -- the superset (adds e.g. ELO soft-temperature, bootstraps) recorded in metadata and replayed by `--rerun`; excludes output folder / verbosity / cache toggle.

This is orthogonal to #66 (which standardizes the result payload shape) and #63 (which defines artifacts like `battles.jsonl`). The descriptor is about run inputs/provenance, not the result contract, so they compose.

### Suggested split:

1. **Descriptor + content-addressed completion cache** (pairwise / MT-Bench / ELO). Small and self-contained; mainly closes the gap where bumping a pinned dataset revision doesn't bust the completion cache. Doesn't touch results JSON or judge caching, so it can land alongside #66 and #63.
2. **Unified judge-completion cache** -- I'd love to fold this into #36 rather than do it separately. One thing I'd flag on the proposed (benchmark, instruction_id, model_a, model_b, judge) key: it doesn't capture the judge sampling params, prompt, truncation or swap mode, so changing temperature/prompt would hit a stale entry. I'd suggest keeping that human-readable composite for row identity/manual deletion, but adding a descriptor/version hash (judge kwargs + prompt hashes + truncation) so those changes invalidate correctly. sqlite is fine; the descriptor just supplies the "what invalidates" function — and it lets us cache judgements in pairwise/MT-Bench too, not just ELO. I'd add this to #36.
3. **run-metadata v2** -- additive, centralized in `repro.write_run_metadata` (a run_descriptor + `run_descriptor_sha256`, the round-trippable `config_resolved`, and a per-artifact sha256). To avoid two schema bumps, I'd rather land this after #63 so it's one coordinated change to the metadata writer instead of each pipeline writing its own.
4. **`--rerun`** -- agree it's mostly redundant with `--config_path` until judge caching exists; its value is faithful replay of resolved settings + reusing cached judgements, so it naturally comes after (2) and (3). I'd ship it last as a thin wrapper that reads the metadata's `config_resolved`.

I have a working prototype of all four on a branch that implements some of the details I explained above! -- https://github.com/OpenEuroLLM/JudgeArena/compare/main...alexrs-cohere:JudgeArena:feat/reproducible-runs

I will be traveling next week and my bandwidth will be limited, but feel free to use the prototype I linked above if you believe it is useful and relevant!

ErlisLushtaku:
@alexrs-cohere, I agree with the two-layer descriptor and splitting into separate PRs. Go ahead opening them whenever you like.

Only one point regarding Judge-completion cache, let's pull this into its own PR and settle the design in #36 first. Two things I think we should have there: (1) per-battle granularity rather than the whole-pass ELO blob, so partial/incremental runs and pairwise/MT-Bench all benefit, (2) what we key on - your descriptor-hash is clean, but since it references completions by hash, not content, a hit is only sound if it is deterministic, otherwise the same descriptor can map to different completions. I am not sure if we can guarantee for every model backend that they will give the same completion given the same descriptor. So let's discuss keying on the judge input vs descriptor-only in #36.

No rush given your travel - we can keep the #36 discussion going meanwhile. Thanks for helping on this!

alexrs-cohere:
@ErlisLushtaku Sounds good! I'll start opening the split PRs (descriptor + content-addressed completion cache first, then run-metadata v2 once #63 lands and I find some time). Agreed that judge-completion caching is its own PR (once we agree on the details in #36)

And you're right about the keying. Keying the judge cache on the completion descriptor hash is only sound if descriptor -> completions is deterministic, which we can't guarantee across backends (temp>0, unseeded, providers that ignore seed, etc). I think the resolution is two layers with different keys: completion cache stays descriptor-keyed (it is the thing that pins completions for a config), but the judge cache keys on the content of the judge input (rendered prompt + A/B text + judge model + sampling kwargs), which is sound regardless of generation determinism. I'll write this up in #36 and we can discuss the details there!

PR #42, opened 2026-05-12
Integration of Soft-ELO

kargibora:
Implements the Soft-Elo pipeline: feed the judge's calibrated score-difference into the Bradley–Terry fit as a soft preference $\tilde y = \sigma(\beta s)$ instead of discretising to win/loss/tie. Optionally MLE-fit  $\beta$ on human-labeled arena battles before the main run.

  ### What changed

  - **`fit_bradley_terry`** (`estimate_elo_ratings.py`) replaces   `compute_bradley_terry`. Takes a soft target `pref_col ∈ [0,1]` (0=A wins,1=B wins, 0.5=tie) and uses the standard soft-CE → weighted-LR decomposition. Hard labels (`{0, 0.5, 1}`) reduce to the previous fit.
  - **Temperature calibration** (`evaluate.calibrate_temperature`): concave MLE for $\beta^\star$ via `scipy.optimize.minimize_scalar` on $\sum\log\sigma(\beta(2y-1)\Delta s)$. Driven from `estimate_elo_ratings.main` — samples human battles, reruns the  judge on them, parses raw scores with `PairScore(temperature=1.0)`, fits  $\beta^\star$, then re-parses all cached judge completions with the calibrated   temperature (handles `swap_mode="both"` reconstruction).
  - **Reporting**: human-only BT ratings computed as ground-truth reference, prints MAE vs Human-Elo on overlapping models; return dict gains   `human_elo`, `mae_vs_human`, `method`, `calibrated_temperature`.

  ### New flags

  | Flag | Default | Effect |
  |---|---|---|
  | `--soft-elo` | off | Use soft BT targets instead of hard {0, 0.5, 1} labels. |
  | `--soft-elo-temperature` | `0.3` | Initial $\beta$; overridden if calibration runs. Empirical range across judges in the paper: `[0.36, 0.60]`. |
  | `--calibrate-temperature` | off | MLE-fit $\beta^\star$ on human-labeled arena battles before the run. Requires `--soft-elo`; warns and skips otherwise. |
  | `--calibration-size` | all human battles | Number of human battles to sample for calibration. Needs `--calibrate-temperature`. |

  ### How to run

  Hard-Elo (unchanged behavior):
  ```bash
  judgearena --task elo-lmarena-100k \
    --model_A Together/meta-llama/Llama-3.3-70B-Instruct-Turbo \
    --judge_model OpenRouter/deepseek/deepseek-chat-v3.1 \
    --n_instructions 200

  Soft-Elo with calibration (recommended):
  judgearena --task elo-lmarena-100k \
    --model_A Together/meta-llama/Llama-3.3-70B-Instruct-Turbo \
    --judge_model OpenRouter/deepseek/deepseek-chat-v3.1 \
    --n_instructions 200 \
    --soft-elo --calibrate-temperature --calibration-size 300
```

### How to test

  uv run pytest tests/test_cli.py tests/test_estimate_elo_ratings.py

-   test_cli.py covers the new flags routing through the unified entrypoint;
-   test_estimate_elo_ratings.py covers fit_bradley_terry and the main pipeline.

### CODE REVIEW

**Comment by kargibora** [comment]
With latest commit [d53cf64](https://github.com/OpenEuroLLM/JudgeArena/pull/42/commits/d53cf6414896c42328a1296ebd57057c25f12750), `--soft-elo` is the default flag (opt `--no-soft-elo` to use normal one). However I did not change the `--calibrate-temperature` as default, because I think this requires some extra **computation**.

**Comment by ErlisLushtaku** [comment]
@kargibora is this ready for review?

**Comment by kargibora** [comment]
> @kargibora is this ready for review?

Yes it is!

File: judgearena/estimate_elo_ratings.py [hunk 1]
```diff
@@ -33,120 +33,142 @@ class CliEloArgs(BaseCliArgs):
...
+                "model_b": opp,
+                "winner": winner,
+                "pref": pref,
+            }
+        else:
+            rec = {
+                "model_a": opp,
+                "model_b": model_name,
+                "winner": winner,
+                "pref": None if _is_nan_pref(pref) else 1.0 - pref,
```
ErlisLushtaku: This inverts the soft preference for every battle where our model is in position B. model_a/model_b here already encode position (A-model / B-model), and pref is P(model_b wins), the same convention fit_bradley_terry uses, so it must be stored unchanged, exactly like the is_pos_a branch on line 158. 

With 1.0 - pref, pref disagrees with both winner and pref_hard, and since soft_elo is now the default, the model under evaluation gets systematically corrupted ratings (a 100%-win model ranks last). I think we should use "pref": pref (keep the None/NaN handling).

judgearena/estimate_elo_ratings.py
```diff
@@ -373,16 +396,153 @@ def run_judge() -> pd.DataFrame:
...
+    score_parser = PairScore(
+        temperature=calibrated_temperature
+        if calibrated_temperature is not None
+        else args.soft_elo_temperature
+    )
+
+    # If we calibrated the temperature, the prefs stored in df_judge were
+    # computed with the default T=0.3.  Re-parse them with the new parser so
+    # the soft-ELO bootstrap uses calibrated preferences.
+    if calibrated_temperature is not None:
```
ErlisLushtaku: score_parser is only used when calibration ran, so --soft-elo-temperature is dead in the default path (cached df_judge["pref"] is always T=0.3, and the cache key doesn't include temperature). Would recommend to always re-parsing from df_judge["judge_completion"] with score_parser for the soft path (drop the if calibrated_temperature is not None gate), which also unifies the duplicated swap_mode="both" recombination logic with the block in run_judge.

judgearena/estimate_elo_ratings.py
```diff
@@ -373,16 +396,153 @@ def run_judge() -> pd.DataFrame:
...
+                max_tokens=args.max_out_tokens_judge,
+                **judge_extra_kwargs,
+            )
+            cal_annotations, _, cal_prefs = judge_and_parse_prefs(
+                judge_chat_model=judge_chat_model_cal,
+                instructions=cal_instructions,
+                completions_A=cal_completions_a,
+                completions_B=cal_completions_b,
+                swap_mode=args.swap_mode,
+                truncate_input_chars=args.truncate_all_input_chars,
```
ErlisLushtaku: Calibration should mirror the main judge run so T* matches the score distribution it's applied to, so I think we should use `truncate_input_chars=args.truncate_judge_input_chars` and pass `provide_explanation=args.provide_explanation`. Otherwise the calibrated temperature is fit on a different prompt/truncation regime than the evaluation.

judgearena/estimate_elo_ratings.py
```diff
@@ -373,16 +396,153 @@ def run_judge() -> pd.DataFrame:
...
+                "--calibrate-temperature has no effect with --no-soft-elo; skipping."
+            )
+        else:
+            logger.info("Calibrating PairScore temperature against human annotations.")
+            # Sample calibration battles from the already-loaded arena battles.
+            # Use the same judge to score them so scores and labels are comparable.
+            _cal_n = (
+                min(args.calibration_size, len(df_arena))
+                if args.calibration_size is not None
+                else len(df_arena)
```
ErlisLushtaku, line 547: Defaulting calibration to all arena battles can mean tens of thousands of (uncached) judge calls, which is a big API cost. We could use a default cap (e.g. a few hundred) and wrapping this judge pass in cache_function_dataframe so reruns don't re-pay.

judgearena/estimate_elo_ratings.py
```diff
@@ -373,16 +396,153 @@ def run_judge() -> pd.DataFrame:
...
+
+            # Build (delta_s, y) pairs from calibration battles.
+            # delta_s = score_A - score_B (raw, using default T=1 to extract scores)
+            raw_parser = PairScore(temperature=1.0)
+            delta_s_cal = []
+            y_cal = []
+            for ann, human_winner in zip(
+                cal_annotations, cal_battles["winner"].tolist(), strict=True
+            ):
+                sa = raw_parser.get_regexp_match(
```
ErlisLushtaku: This duplicates PairScore.parse_model_raw's extraction while being slightly different, so they can drift. We could add a PairScore.parse_raw_scores(completion) and call it from both parse_model_raw and here. 

Also raw_parser = PairScore(temperature=1.0) / "using default T=1" (on [this line](https://github.com/OpenEuroLLM/JudgeArena/pull/42/changes#diff-9fee7a3e27c424d4de0a0d4570ed7c81e3f00c1512199ba212b03df4577094f3R465)) could be misleading as get_regexp_match doesn't use temperature.

judgearena/estimate_elo_ratings.py
kargibora: This is something we have discussed with David. We agreed that the default parameters should be the ones that works _best_ according to us, but I am more biased towards making this optional as you have suggested. @geoalgo

judgearena/estimate_elo_ratings.py
kargibora: Oh you are right, thanks for caching it.

judgearena/estimate_elo_ratings.py
kargibora: [734c08b](https://github.com/OpenEuroLLM/JudgeArena/pull/42/commits/734c08b5409c591b8f78a812034440d2e2cf0f67) fixes this, thanks!

judgearena/estimate_elo_ratings.py
kargibora: Closed with [7411407](https://github.com/OpenEuroLLM/JudgeArena/pull/42/commits/7411407071f5225dbe880b43fabcec7f57f15281)

kargibora:
I have pushed the changes, thank you @ErlisLushtaku! It will be ready to merge after deciding whether to set soft-elo default or not. I think discussing this is important again.

ErlisLushtaku (approved):

PR #48, opened 2026-05-19
2. Add prompt presets and mt-bench changes

ErlisLushtaku:
## Summary
- Move shared judge prompt preset resolution into its own stack layer.
- Add Skywork and localized m-Arena-Hard v2 prompt presets/assets.
- Wire prompt preset selection through CLI, pairwise judging, and MT-Bench preset judging.

## Stack
- Base: `pr32-split/01-judge-limits-baselines` (#43).
- Next: `pr32-split-v2/03-inference-metadata-thinking`.

## Test plan
- `uv run pytest tests/test_cli.py tests/test_regexp.py tests/test_mt_bench_downloads.py tests/test_generate_and_evaluate.py`
- `uv run ruff check .`

Includes-AI-Code: true


judgearena/evaluate.py
kargibora, line 1: This is much cleaner than what we have now! Amazing

judgearena/judge_prompt_presets.py
```diff
@@ -0,0 +1,173 @@
...
+
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Literal
+
+JudgeParserMode = Literal["score", "verdict"]
+
+DEFAULT_JUDGE_PROMPT_PRESET = "default"
+SKYWORK_JUDGE_PROMPT_PRESET = "skywork"
+M_ARENA_HARD_V2_LOCALIZED_PROMPT_PRESETS = {
```
kargibora: I don't know this but is m-arena-hard only restricted to these four languages? Or this is the subset we are using?

kargibora (approved):
I see no apperant problems. Quite well written and does not introduce many complications

judgearena/judge_prompt_presets.py
ErlisLushtaku: No it has more languages, these are only the ones we ship translations for.

geoalgo:
Just had a quick chat on this, we probably want to move the localized prompts in a separate package for the paper experimetns and also skyworks setup.

ErlisLushtaku:
closing as this was split into #54 and #55

PR #50, opened 2026-05-19
5. Add Elo sampling and Gemini safety handling

ErlisLushtaku:
## Summary
- Add Elo sampling controls and CLI wiring.
- Add OpenRouter Gemini safety-tolerant chat handling.
- Keep this PR on top of the reordered prompt and inference stack.

## Stack
- Base: `pr32-split-v2/03-inference-metadata-thinking`.
- Next: `pr32-split-v2/05-paper-experiment-scripts`.
- Replaces old #46 on the reordered stack.

## Test plan
- `uv run pytest tests/test_estimate_elo_ratings.py tests/test_cli.py tests/test_utils.py`
- `uv run pytest tests`
- `uv run ruff check .`

### DIFF
--- a/judgearena/estimate_elo_ratings.py
+++ b/judgearena/estimate_elo_ratings.py
@@ -1,6 +1,9 @@
 import hashlib
+import json
+import re
 from dataclasses import dataclass
 from functools import partial
+from pathlib import Path
 
 import numpy as np
 import pandas as pd
@@ -33,6 +36,7 @@ class CliEloArgs(BaseCliArgs):
     n_bootstraps: int = 20
     seed: int = 0
     baseline_model: str | None = None
+    elo_random_battles: int | None = None
 
 
 def compute_bradley_terry(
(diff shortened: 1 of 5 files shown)


judgearena/estimate_elo_ratings.py
```diff
@@ -159,17 +260,34 @@ def main(args: CliEloArgs) -> dict:
...
+                .head(args.n_instructions_per_language)
+                .reset_index(drop=True)
+            )
 
-    # Keep at most n_instructions total (subset used for LLM-judge evaluation)
-    if args.n_instructions is not None:
-        df_battles = df_battles.head(args.n_instructions)
+        # Keep at most n_instructions total (subset used for LLM-judge evaluation)
+        if args.n_instructions is not None:
+            df_battles = df_battles.head(args.n_instructions)
```
kargibora, line 290: This is possibly deterministic unless we shuffle the rows. Is this expected? Do we need some seed + shuffling + sampling strategy?

kargibora (commented):
Much smaller and quick PR. I don't see any possible problems but do we want to pass `REFUSAL_MARKER` to the judge? Is there any advantage of recording these instances instead of throwing an error for the battle?

geoalgo:
Just discussed it, we probably want to move that to the paper branch. We may also want to handle safety rejects from the judge but then we would want something more generic that works for all/most models.

judgearena/estimate_elo_ratings.py
ErlisLushtaku: I am not sure, we'd have to discuss with David, but in any case we could do that later in another PR.

PR #51, opened 2026-05-19
6. Add paper experiment scripts

ErlisLushtaku:
## Summary
- Add paper-phase analysis and runner scripts.
- Update the benchmark launcher default judge model wiring.
- Keep paper reproduction scripts at the top of the reordered stack.

## Stack
- Base: `pr32-split-v2/04-elo-sampling-gemini`.
- Replaces old #47 on the reordered stack.

## Test plan
- `uv run python -m py_compile scripts/analyze_lmarena_elo.py scripts/phase_b_marena_localized_prompt_table.py scripts/phase_b_paper_table.py scripts/phase_b_serial_runner.py slurmpilot_scripts/launch_generation_and_evaluation.py`
- `uv run pytest tests`
- `uv run ruff check .`

Includes-AI-Code: true

### DIFF
--- a/judgearena/generate.py
+++ b/judgearena/generate.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from typing import Any
 
 import pandas as pd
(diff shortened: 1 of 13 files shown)


kargibora (commented):
Thanks for the PR!

I know that we want to put scripts for reproducilibity, but I think this should not be in the main branch but should stay on a different branch like `paper`.  Otherwise, it would be confusing for the end users who want to use our framework for running/evaluating their models on dynamic benchmarks.

geoalgo:
Just discussed it, will move to the paper branch to reproduce the results.

PR #52, opened 2026-05-19
3. Metadata and truncation tracking

ErlisLushtaku:
## Summary
- Propagate generation metadata end-to-end so `finish_reason`/`stop_reason` survive both async and batch inference paths instead of being dropped before result writing.
- Add structured limit-event tracking for generation and judging, including char truncation, judge-input token-budget preflight shrinking, and generation-only `--skip_judging` runs that still write useful summaries.
- Surface those diagnostics consistently in saved results for pairwise benchmarks and MT-Bench so long-output and context-limit failures become visible instead of silent.

## Test plan
- `uv run pytest tests/test_utils.py tests/test_regexp.py tests/test_generate_and_evaluate.py tests/test_mt_bench_downloads.py`

### DIFF
--- a/judgearena/cli_common.py
+++ b/judgearena/cli_common.py
@@ -28,6 +28,7 @@ class BaseCliArgs:
     ignore_cache: bool = False
     judge_prompt_preset: str = "default"
     mt_bench_judge_mode: str = "default"
+    skip_judging: bool = False
     truncate_all_input_chars: int = 8192
     truncate_judge_input_chars: int | None = None
     max_out_tokens_models: int = 32768
@@ -54,6 +55,17 @@ class BaseCliArgs:
         )
 
 
+def parse_optional_bool(raw: str | None) -> bool:
+    if raw is None:
+        return True
+    normalized = raw.strip().lower()
+    if normalized in {"1", "true", "yes", "on"}:
+        return True
+    if normalized in {"0", "false", "no", "off"}:
+        return False
+    raise argparse.ArgumentTypeError(f"Expected a boolean value, got '{raw}'.")
+
+
 def add_common_arguments(parser: argparse.ArgumentParser) -> None:
     """Register the CLI flags shared by all judgearena entrypoints."""
     parser.add_argument(
@@ -75,7 +87,10 @@ def add_common_arguments(parser: argparse.ArgumentParser) -> None:
     )
     parser.add_argument(
         "--provide_explanation",
-        action="store_true",
+        nargs="?",
+        const=True,
+        default=False,
+        type=parse_optional_bool,
         help=(
             "If specified, judge will provide explanation before making a "
             "judgement. Does not necessarily improve the accuracy of the judge "
@@ -96,7 +111,10 @@ def add_common_arguments(parser: argparse.ArgumentParser) -> None:
     )
     parser.add_argument(
         "--ignore_cache",
-        action="store_true",
+        nargs="?",
+        const=True,
+        default=False,
+        type=parse_optional_bool,
         help="If specified, ignore cache of previous completions.",
     )
     parser.add_argument(
@@ -106,7 +124,8 @@ def add_common_arguments(parser: argparse.ArgumentParser) -> None:
         default="default",
         help=(
             "Judge prompt preset to use. 'default' preserves the existing score-first "
-            "JudgeArena prompts, while 'skywork' enables a verdict-first preset."
+            "JudgeArena prompts, while 'skywork' enables an optional Skywork-style "
+            "verdict-first preset."
         ),
     )
     parser.add_argument(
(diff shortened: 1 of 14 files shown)


judgearena/generate_and_evaluate.py
```diff
@@ -303,61 +422,58 @@ def main(args: CliArgs):
         baseline_plan.display_name,
     )
 
-    # TODO currently we just support base models for fluency, we could also support instruction-tuned models
```
ErlisLushtaku, line 306: Should this comment have been removed?

judgearena/mt_bench/common.py
```diff
@@ -5,7 +5,7 @@
 
 import pandas as pd
 
-from judgearena.utils import safe_text
+from judgearena.utils import safe_text_with_metadata
 
 MT_BENCH_REFERENCE_CATEGORIES: set[str] = {
```
kargibora, line 10: Do we really need a seperate fields for _truncated_ prompts?

judgearena/utils.py
```diff
@@ -33,6 +36,123 @@ def _data_root_path() -> Path:
...
+    stage: str
+    field: str | None = None
+    case_id: str | None = None
+    model_spec: str | None = None
+    original_length: int | None = None
+    final_length: int | None = None
+    note: str | None = None
+
+
+class LimitEventTracker:
```
kargibora, line 93: What is the purpose of this class? Could you eloborate it (or possibly some docstring would be helpful

How many events we have? Does it only records the pre/post filterings or transformations for reproducibility?

judgearena/utils.py
```diff
@@ -113,6 +233,33 @@ def truncate(s: str, max_len: int | None = None) -> str:
...
+    tracker: LimitEventTracker | None = None,
+    kind: str | None = None,
+    stage: str | None = None,
+    field: str | None = None,
+    case_id: object | None = None,
+    model_spec: str | None = None,
+) -> tuple[str, bool]:
+    original = s if isinstance(s, str) else ""
+    truncated = truncate(original, max_len=max_len)
+    was_truncated = truncated != original
```
kargibora, line 249: Not super important as string matchings are quite fast, but do we really need it? 

We can simply say 
```py
if len(original) < max_len:
# do not truncate
else:
# truncate
```

judgearena/utils.py
```diff
@@ -127,17 +274,75 @@ def safe_text(value: object, truncate_chars: int | None) -> str:
...
     if use_tqdm:
         # perform inference asynchronously to be able to update tqdm, chat_model.batch does not work as it blocks until
         # all requests are received
+        # JUDGEARENA_JUDGE_MAX_CONCURRENCY caps simultaneous in-flight ainvokes
+        # (e.g. against OpenRouter). Unset = unbounded, preserving prior behaviour.
+        cap_raw = os.environ.get("JUDGEARENA_JUDGE_MAX_CONCURRENCY")
+        cap = int(cap_raw) if cap_raw and int(cap_raw) > 0 else None
+
         async def process_with_real_progress(chat_model, inputs, pbar):
+            sem = asyncio.Semaphore(cap) if cap else None
```
kargibora, line 344: That is great but:
* **(1)** Are we sure there is no concurrent writing/reading that can create problems?  There should be no race condition or things can go messy
* **(2)** Assuming we want to use local machines instead of OpenRouter, does this code also spawns multiple agents or we will only use it for OpenRouter/API based services.

judgearena/utils.py
```diff
@@ -221,6 +454,52 @@ async def ainvoke(self, input, **invoke_kwargs):
         return self.message
 
 
+_VLLM_INIT_RETRY_SIGNATURES = (
```
kargibora, line 457: Defining this constant here feels wrong. Maybe we should have `constants.py` and import it from here, otherwise hard to edit or read.

judgearena/cli_common.py
```diff
@@ -75,7 +87,10 @@ def add_common_arguments(parser: argparse.ArgumentParser) -> None:
     )
     parser.add_argument(
```
kargibora, line 88: I think it is quite straightforward to use store_true. Do we really need to check the edge cases like user providing
`--provide_explanation "yes"`

judgearena/evaluate.py
```diff
@@ -22,14 +23,22 @@
...
     download_hf,
     read_df,
-    truncate,
+    truncate_with_metadata,
 )
 
+if TYPE_CHECKING:
+    from transformers import PreTrainedTokenizerBase
+
+_PREFLIGHT_MAX_ITERATIONS = 3
```
kargibora, line 38: Repeated constants, we can unify such definitions in a seperate file

judgearena/evaluate.py
```diff
@@ -448,3 +553,161 @@ def _none_to_nan(x):
         prefs = pd.concat([prefs, (1 - prefs_reversed)]).reset_index(drop=True)
 
     return annotations, annotations_reversed, prefs
+
+
+_LC_ROLE_MAP = {"human": "user", "ai": "assistant", "system": "system"}
+
+
+def _count_chat_tokens(prompt_value: Any, tokenizer: Any) -> int:
```
kargibora, line 561: We shouldn't put these utilitiy functions here. Larger the `evaluate.py` is, harder to add more features in the future.  Filterins and resolvers should be defined seperately and imported.

kargibora (commented):
Thanks for the PR!

I have several recommendations. 

I feel like we are changing too many things at once. There are lots of fields, `args` and resolving these to defaults as a code. This can be simplified and unified: for example we can create an utils folder, move dataset utils there, constants (it can be its own file as well), utils for evaluate generate... This allow us to also "unify" possible functions together

geoalgo:
Just discussed it, closing for now, we may want to readd it later but we will do that once other PRs are merged in.

PR #54, opened 2026-05-28
2. Add prompt presets

ErlisLushtaku:
## Summary
- Move shared judge prompt preset resolution into its own stack layer.
- Wire prompt preset selection through CLI, pairwise judging, and MT-Bench preset judging.

## Stack
- Next: `pr32-split-v2/02.5-mt-bench-preset-judging`.


ErlisLushtaku:
@kargibora this is the same as #48 I just split some logic related to mt-bench into another one #55 . If you want you can give it another review but it should be the same as what you saw before. #55 is new though.

judgearena/prompts/registry.py
```diff
@@ -0,0 +1,249 @@
...
+        )
+    if system_file is not None and user_file is not None:
+        return _resolve_file_prompt(
+            system_file=system_file,
+            user_file=user_file,
+            multi_turn=multi_turn,
+            provide_explanation=provide_explanation,
+        )
+
+    if preset is None:
```
kargibora, line 187: If `preset=None`, we set it to `DEFAULT_WITH_EXPLANATION_PRESET` if explanation is provided. So we can not use `task` and `provide_explanation` at the same time. Is this intended?

judgearena/prompts/registry.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+def default_preset_for_task(task: str | None) -> str:
+    if task is None:
+        return DEFAULT_JUDGE_PROMPT_PRESET
+    if task in TASK_DEFAULT_PRESET:
+        return TASK_DEFAULT_PRESET[task]
+    if task.startswith("m-arena-hard"):
+        return DEFAULT_JUDGE_PROMPT_PRESET
+    if task.startswith("fluency"):
```
kargibora, line 110: I think checking tasks specifically for these keywords are not so optimal.  If we use `*-fleuncy* it will fail for example. We can use asserts and task resolving structure for this one.

judgearena/mt_bench/preset_judging.py
```diff
@@ -0,0 +1,365 @@
...
+    if swap_mode == "both":
+        swapped_judgments, swapped_prompt_kwargs = _infer_by_prompt_groups(
+            judge_chat_model=judge_chat_model,
+            items=items,
+            use_tqdm=use_tqdm,
+            swap_answers=True,
+        )
+        _append_results(swapped_judgments, swapped_prompt_kwargs, swapped=True)
+
+    return pd.Series(preferences, dtype=float), annotations, metadata, 0
```
kargibora: It is a bit confusing, we say return type is `None` but we return tuple with 4 elements?

judgearena/mt_bench/mt_bench_utils.py
```diff
@@ -151,14 +161,69 @@ def _run_mt_bench_fastchat(
...
+    args: CliArgs,
+    res_folder: Path,
+    result_name: str,
+    questions_df: pd.DataFrame,
+    completions_a: pd.DataFrame,
+    completions_b: pd.DataFrame,
+    judge_chat_model,
+    resolved_prompt: ResolvedJudgePrompt,
+) -> pd.Series:
+    prefs, annotations, combined_metadata, _num_inconsistent = (
```
kargibora: We return 0 in the `judge_mt_bench_with_preset`, even if we do not, we never use `_num_inconsistent` variable. We should either discard it ore use it properly.

kargibora (commented):
Thanks for the clean-up! I left some more comments after reading it throughly.

judgearena/prompts/registry.py
ErlisLushtaku: At the moment we don't have any other tasks that don't fit in this. Maybe we can leave this for the future rather than engineering it prematurely. Wdyt?

judgearena/prompts/registry.py
ErlisLushtaku: This precedence is unchanged from the original judge-prompt-registry PR (#40), where provide_explanation was intentionally documented as a legacy alias for default_with_explanation that overrides the task default. I also added a comment [here](https://github.com/OpenEuroLLM/JudgeArena/pull/54/changes/50531a54d59e61d005e25551da138f8914405356#diff-6ead50e4270e50b5883f01451735759857284f0dd2d1687eed5aa84a28d234a0R188-R191) to make this more clear.

judgearena/mt_bench/preset_judging.py
ErlisLushtaku: Agreed, that trailing 0 was unnecessary. It only existed to mirror the FastChat path's num_inconsistent, which preset judging doesn't compute. Removed it [here](https://github.com/OpenEuroLLM/JudgeArena/pull/54/changes/50531a54d59e61d005e25551da138f8914405356#diff-6ead50e4270e50b5883f01451735759857284f0dd2d1687eed5aa84a28d234a0R188-R191).

judgearena/mt_bench/mt_bench_utils.py
ErlisLushtaku: Check [this comment](https://github.com/OpenEuroLLM/JudgeArena/pull/54#discussion_r3412175126)

judgearena/prompts/registry.py
kargibora: Ofc I agree

judgearena/prompts/registry.py
kargibora: In that case, do we really need to keep legacy alias/arguments to not break previous CLI commands?  I think this is an overall design question but I think it is more natural to just drop alias in the functions and just create a warning about the usage

judgearena/mt_bench/preset_judging.py
kargibora: Thanks!

judgearena/mt_bench/mt_bench_utils.py
kargibora: Thanks

judgearena/prompts/registry.py
ErlisLushtaku: It might be a point of discussion, we could discuss that in a meeting later

ErlisLushtaku: I'll note this, and merge the PR for now so we don't block further and have to resolve more conflicts

PR #55, opened 2026-05-28
2.5. mt-bench changes

ErlisLushtaku:
## Summary

This PR pulls the MT-Bench-specific follow-up work out of the prompt preset PR and keeps it as a focused stacked cleanup.

It now does two things:

- refactors the non-delegated MT-Bench judging path to reuse shared pairwise helpers instead of duplicating batching, prompt grouping, answer swapping, and item construction
- adds MT-Bench-specific reproducibility/orchestration follow-ups that were originally mixed into PR 52, without bringing over the deferred truncation tracking or cache-key changes

Concretely, this PR:

- extracts shared MT-Bench pairwise judging helpers into `judgearena/mt_bench/pairwise_judging.py`
- keeps FastChat-specific verdict parsing and preset-specific score parsing separate, while reusing the same lower-level MT-Bench mechanics
- replaces loose shared item dicts with a typed `MTBenchJudgeItem`
- makes the two `swap_mode="both"` semantics explicit in code:
  - FastChat: conservative agreement
  - preset judging: append inverted swapped scores
- factors shared MT-Bench result finalization in `mt_bench_utils.py`
- aligns pre-generated or cached MT-Bench completions to the requested question order and raises a clear error if rows are missing
- writes MT-Bench run metadata, including input payload and prompt metadata, alongside saved artifacts
- lets `run_mt_bench` derive its default result folder and artifact name when callers do not pass them explicitly
- adds focused MT-Bench tests for both the FastChat and preset judging paths, plus coverage for completion alignment and metadata writing


judgearena/mt_bench/preset_judging.py
```diff
@@ -311,11 +187,14 @@ def _append_results(
         used_prompt_kwargs: list[dict[str, str]],
         *,
         swapped: bool,
+        swap_policy: MTBenchSwapPolicy = _PRESET_SWAP_POLICY,
```
kargibora: Is this parameter configurable? It is always constant right?

judgearena/mt_bench/mt_bench_utils.py
```diff
@@ -224,6 +305,7 @@ def run_mt_bench(
     result_name: str,
 ):
     """MT-Bench pipeline with preset or FastChat-original pairwise judging."""
+    run_started_at = datetime.now(UTC)
```
kargibora, line 306: Maybe use `datetime.now(UTC).isoformat()`.  We also record a timeframe as:
```py
        "date": str(datetime.now().isoformat()),
```
above.

judgearena/mt_bench/preset_judging.py
ErlisLushtaku: You're right, it's always constant. It's an internal no-op guard. Since it was constant on both sides (preset_judging and fastchat), I removed it from both sides, and documented the two swap_mode="both" strategies with inline comments instead i.e. conservative agreement for FastChat (winner only if both orderings agree, else tie), append-inverted-score for presets (_normalize_preference(..., swapped=True)).

judgearena/mt_bench/mt_bench_utils.py
ErlisLushtaku: Good point, changed it to be consistent [here](https://github.com/OpenEuroLLM/JudgeArena/pull/55/commits/2c65679aa4dbc994babcbc0543006b61f734deb5).

kargibora: Thank you

PR #57, opened 2026-06-01
3.5. Add runtime robustness, judge-kwarg defaults, CLI/orchestration cleanup, point to judgearena hf

ErlisLushtaku:
## Summary

Runtime, CLI, and orchestration follow-ups split out of the closed PR #52. This intentionally excludes truncation/limit tracking, generation cache-key hashing, `skip_judging`, and the MT-Bench-specific changes (now in PR #55).

## Changes

- **Runtime / model robustness** (`utils.py`)
  - `_init_llm_with_retry`: retry vLLM construction on transient GPU-init signatures (e.g. `cudaErrorDevicesUnavailable`) with bounded backoff; reraise everything else immediately.
  - `make_model`: strip all vLLM-only engine kwargs (`max_model_len`, `gpu_memory_utilization`, `tensor_parallel_size`, `kv_cache_dtype`, …) for non-vLLM providers so remote APIs don't reject them.
  - `ChatVLLM`: configurable `temperature`/`top_p` with `set_temperature`, exposed tokenizer, and `chat_template_kwargs` plumbing.
- **Metadata propagation bugfix** (`utils.py`)
  - `_extract_ai_message_metadata` + `do_inference(return_metadata=...)` and `ChatVLLM.batch_with_metadata` so `finish_reason`/`stop_reason` are preserved on both the async (`use_tqdm`) and batch paths.
  - Optional `JUDGEARENA_JUDGE_MAX_CONCURRENCY` cap on in-flight async invokes (unset = unbounded, prior behavior).
- **Judge engine defaults** (`utils.py`, `generate_and_evaluate.py`)
  - `build_default_judge_model_kwargs` centralizes judge engine kwargs and adds an FP8 KV-cache default for FP8 judges; `_build_judge_engine_kwargs` now delegates to it.
- **CLI compatibility** (`cli.py`)
  - Restore the deprecated `--model` alias, collapsed into `--model_A` via `_resolve_model_a` (errors if both are given).
- **Generation / evaluation orchestration** (`generate_and_evaluate.py`, `evaluate.py`)
  - Replace `functools.partial` generation wiring with a single `_run_generation` helper; preserve the fluency instruction index; pass an explicit `parser_mode` into `judge_and_parse_prefs`.
- **Dataset/download maintenance** (`utils.py`)
  - `download_hf` points at `judge-arena/judge-arena-dataset`; `download_all` is driven by `M_ARENA_HARD_BASELINES`.
- **Tests**: metadata extraction, async/batch `finish_reason` propagation, vLLM init retry behavior, expanded `make_model` kwarg stripping, and baseline-plan resolution cases.

---
_Supersedes #56 (closed): branch renamed `02-prompt-presets-localized` → `03.5-runtime-cli-followups` and rebased onto the PR #55 tip so it forms a clean linear stack. Same single-commit diff (6 runtime files)._


judgearena/cli.py
```diff
@@ -74,6 +74,10 @@ def _build_parser() -> argparse.ArgumentParser:
         "--model_B",
         help="Model B for generate+judge tasks (not yet supported for elo tasks).",
     )
+    parser.add_argument(
```
geoalgo: I am just skimming through this PR and I was wondering about this.
I see the point of trying to have another flag, but isnt it better in this case since this not really model_A as the position are randomized and it may be confusing to users?

judgearena/cli.py
```diff
@@ -152,6 +156,22 @@ def _resolve_task(args: argparse.Namespace) -> str:
     return f"{ELO_TASK_PREFIX}{lower_arena}"
 
 
+def _resolve_model_a(args: argparse.Namespace) -> str | None:
```
kargibora: Creating alias is a bit confusing to me. We should always be consistent, even if it breaks some previous codes no need to bring that back.  More arguments (althoug more configurability) results in harder code.

judgearena/generate_and_evaluate.py
```diff
@@ -191,9 +195,11 @@ def _resolve_baseline_plan(
 
 
 def _build_judge_engine_kwargs(args: CliArgs) -> dict[str, object]:
```
kargibora: I know this is a mirror function but do we really need it?

judgearena/utils.py
```diff
@@ -179,9 +246,26 @@ def batch_with_retry(batch_inputs, max_retries=5, base_delay=1.0):
                 ]
                 try:
                     results = []
+                    results_metadata = []
                     for chunk in chunks:
-                        results.extend(chat_model.batch(inputs=chunk, **invoke_kwargs))
-                    return results
+                        if return_metadata and hasattr(
```
kargibora: We actually never pass `return_metadata=True`. We only pass it for the tests. Can you check this behaviour? 

Also it would be nice to have some docstring for non-local functions.

judgearena/utils.py
```diff
@@ -221,6 +307,43 @@ async def ainvoke(self, input, **invoke_kwargs):
         return self.message
 
 
+_VLLM_INIT_RETRY_SIGNATURES = (
+    "cudaErrorDevicesUnavailable",
+    "CUDA-capable device(s) is/are busy or unavailable",
+    "CUDA error: initialization error",
+)
+_VLLM_INIT_MAX_ATTEMPTS = int(os.getenv("JUDGEARENA_VLLM_INIT_MAX_ATTEMPTS", "4"))
+_VLLM_INIT_BACKOFF_SECONDS = int(
```
kargibora: Possibly if we set `JUDGEARENA_VLLM_INIT_BACKOFF_SECONDS=" "`, then the entire CLI crashes as this is loaded before any functionality. Top-level functionality should never do these type of conversions without checking it. (If the input is malformed, better use default one or crash with a proper error logging)

judgearena/utils.py
```diff
@@ -127,17 +163,41 @@ def safe_text(value: object, truncate_chars: int | None) -> str:
...
         # "max_tokens": 100,
     }
+    metadata: list[dict[str, Any]] | None = None
     if use_tqdm:
         # perform inference asynchronously to be able to update tqdm, chat_model.batch does not work as it blocks until
         # all requests are received
+        # JUDGEARENA_JUDGE_MAX_CONCURRENCY caps simultaneous in-flight ainvokes
+        # (e.g. against OpenRouter). Unset = unbounded, preserving prior behaviour.
+        cap_raw = os.environ.get("JUDGEARENA_JUDGE_MAX_CONCURRENCY")
+        cap = int(cap_raw) if cap_raw and int(cap_raw) > 0 else None
```
kargibora: Similar to problem above, `int(cap_raw)` is uncaught.

judgearena/utils.py
geoalgo: We could implement something like `def safe_parse_int(env_var: str) -> int | None`.

judgearena/cli.py
ErlisLushtaku: Removed the `--model` alias entirely for the moment as it isn't on main. 

On the broader concern that `--model_A` reads as confusing given positions are randomized, I agree, but renaming the primary flag touches main and the other split PRs, so I'd rather handle that as a separate cross-cutting rename than in this PR.

judgearena/cli.py
ErlisLushtaku: Good point. Removed the `--model` alias entirely for the moment as it isn't on main. I had added it for the benchmarks where we only have one model, so `model_A` didn't read naturally when the user doesn't specify a `model_B`. That said, it didn't add any functionality. We can think about this in the future, as there is also the issue of position randomization that David mentioned.

judgearena/generate_and_evaluate.py
ErlisLushtaku: Not really needed, removed it.

judgearena/utils.py
ErlisLushtaku: Good catch, seems this was only used for the truncation related code. Removed it from this PR.

judgearena/utils.py
ErlisLushtaku: Implemented that in the [last commit](https://github.com/OpenEuroLLM/JudgeArena/pull/57/commits/eebb5f93cadb604256525fa11a4bf3c511595468)

judgearena/utils.py
ErlisLushtaku: See the [comment above](https://github.com/OpenEuroLLM/JudgeArena/pull/57#discussion_r3413139614).

judgearena/cli.py
kargibora: Yes, I have also an issue for refactoring where we should discuss the argument names. Lets discuss this more detailly later

ErlisLushtaku:
@kargibora @geoalgo is this fine to merge too?

kargibora:
> @kargibora @geoalgo is this fine to merge too?

Yes we can merge this as well

Issue #60, opened 2026-06-04
Quality of life improvements


geoalgo:
Here are a few things that we should fix:
- [ ] **Cache judgements too** -- we currently cache completions but not judgements, [code](https://github.com/OpenEuroLLM/JudgeArena/blob/main/judgearena/generate_and_evaluate.py#L405) which feels inconsistent and also does not allow to easily see judgements for the user
- [x] **Replace `huggingface-cli` with `hf` in the README** — the README download commands trigger a deprecation warning; `hf download` is the current API.
- [x] **Warn on unparseable judge outputs** — log a warning and show parsed/total counts in the summary table (when a judge fail to parse an output, the battle is dropped and not reported in the count stats which is not ideal).
- [x] **Document `chat_format` for LlamaCpp example** — the llama-2 fallback is silently incorrect for Gemma and Llama-3.x models.
- [x] **Swap_mode="both" reports doubled num_battles** — generate_and_evaluate.py:312 prefs from judge_and_parse_prefs with swap_mode="both" has length 2*n. compute_pref_summary counts num_battles = len(prefs), so the results display shows twice the actual number of unique instructions evaluated.
- [ ] **Provide pre-cached baselines for m-arena-hard-v0.1** — or at minimum document that `CohereLabs/aya-expanse-8b` is not generatable locally and requires a remote API.
- [x] **Report path in result summary** so that user can easily see completions
- [x] **Bloated dependencies** with seaborn / ipython / etc which are not used and where added unintentionally

geoalgo:
Closing as we have an issue tracking caching.

PR #61, opened 2026-06-04
Various fixes and small improvements

geoalgo:
Solving most issues mentioned in https://github.com/OpenEuroLLM/JudgeArena/issues/60

### DIFF
--- a/judgearena/generate_and_evaluate.py
+++ b/judgearena/generate_and_evaluate.py
@@ -13,7 +13,11 @@ from pathlib import Path
 import pandas as pd
 
 from judgearena.cli_common import BaseCliArgs
-from judgearena.evaluate import judge_and_parse_prefs, resolve_judge_prompts
+from judgearena.evaluate import (
+    JudgeAnnotation,
+    judge_and_parse_prefs,
+    resolve_judge_prompts,
+)
 from judgearena.generate import generate_base, generate_instructions
 from judgearena.instruction_dataset import load_instructions
 from judgearena.instruction_dataset.arena_hard import (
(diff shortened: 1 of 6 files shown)


judgearena/generate_and_evaluate.py
```diff
@@ -402,18 +416,86 @@ def _load_or_generate_completions(model_spec: str) -> pd.Series:
...
-        instructions=instructions.head(n_instructions).tolist(),
-        completions_A=completions_A.head(n_instructions).tolist(),
-        completions_B=completions_B.head(n_instructions).tolist(),
-        swap_mode=args.swap_mode,
-        provide_explanation=args.provide_explanation,
-        system_prompt=effective_judge_system_prompt,
-        user_prompt_template=judge_user_prompt_template,
-        truncate_input_chars=args.truncate_judge_input_chars,
-        use_tqdm=args.use_tqdm,
+    judge_cache_name = f"judge_{name}_{n_instructions}"
```
geoalgo: Note that it would be better to refactor `judge_and_parse_prefs` to return a cachable dataframe rather than having to construct preferences twice but it could be done in a follow-up PR.

judgearena/generate_and_evaluate.py
```diff
@@ -402,18 +421,87 @@ def _load_or_generate_completions(model_spec: str) -> pd.Series:
...
+    def _df_to_annotations(df: pd.DataFrame) -> list[JudgeAnnotation]:
+        return [
+            JudgeAnnotation(
+                instruction=row.instruction,
+                completion_A=row.completion_A,
+                completion_B=row.completion_B,
+                judge_completion=row.judge_completion,
+            )
+            for _, row in df.iterrows()
+        ]
```
ErlisLushtaku: The cache doesn't persist judge_input, so the regenerated *-annotations.csv loses the rendered judge prompt column that the non-cached path produced. Is this intentional?

judgearena/generate_and_evaluate.py
```diff
@@ -402,18 +421,87 @@ def _load_or_generate_completions(model_spec: str) -> pd.Series:
...
-        instructions=instructions.head(n_instructions).tolist(),
-        completions_A=completions_A.head(n_instructions).tolist(),
-        completions_B=completions_B.head(n_instructions).tolist(),
-        swap_mode=args.swap_mode,
-        provide_explanation=args.provide_explanation,
-        system_prompt=effective_judge_system_prompt,
-        user_prompt_template=judge_user_prompt_template,
-        truncate_input_chars=args.truncate_judge_input_chars,
-        use_tqdm=args.use_tqdm,
+    judge_cache_name = f"judge_{name}_{n_instructions}"
```
ErlisLushtaku: This cache key omits `provide_explanation` and `truncate_judge_input_chars` (and the judge token/context limits), which directly change judge output. We could also address this later based on the issue #36 .

judgearena/generate_and_evaluate.py
geoalgo: Good point.
Now that I am thinking about this, the way we did it in our paper is to have a mysql cache which is probably cleaner (discussed in https://github.com/OpenEuroLLM/JudgeArena/issues/36). 

I will remove this feature until we have this cache as otherwise we would have backward incompatible cache.

ErlisLushtaku (approved):

PR #63, opened 2026-06-10
Save arena battles and ELO ratings as artifacts

kargibora:
## Problem

The arena ELO pipeline (`estimate_elo_ratings.main`) computed Bradley-Terry ratings but saved nothing usable.
The ratings were only `print`ed and returned (the CLI discards the return value), and the per-battle data lived only in the internal `cache/elo/*.csv.zip` recompute-skipping layer.
There was no battle file and no ratings file, so a run could not be visualized or its ELO recomputed without rerunning the whole GPU job.
The pairwise pipeline already writes a results folder; the arena pipeline did not.

## Solution

Persist the run as a small set of artifacts, built around the battle as the atomic unit: ELO is a pure function of a list of battles, so saving the battles (plus the bootstrap ratings) is enough to reconstruct or re-analyse a run.

New module `judgearena/battles.py`:
- `Battle` — one outcome (`model_a`, `model_b`, `winner`, `source`, `question_id?`, `judge_model?`). `from_dict` ignores unknown keys, so old files keep loading after the schema grows.
- `write_battles` / `read_battles` / `battles_to_frame` — JSONL round-trip plus the bridge to the existing `compute_bradley_terry`.
- `RatingEntry` / `EloReport` — the leaderboard (mean + bootstrap CI per model) and run metadata.

`estimate_elo_ratings.main` now writes a results folder:

```
results/{arena}-{model}-{judge}-{ts}/
  battles.jsonl          # source of truth: every battle, tagged source = llm-judge | human
  elo_ratings.json       # leaderboard: rating + ci_low/ci_high + n_battles per model
  bootstrap_ratings.csv  # full bootstrap matrix, so CIs can be re-plotted without rerunning
  run-metadata.v1.json   # git hash, deps, timings, artifacts manifest (repro.py)
```

The ELO math is unchanged; the combined battle frame just gained `source` / `question_id` / `judge_model` columns that `compute_bradley_terry` ignores.

### Design notes

- `battles.jsonl` inlines the human arena battles (not just the judged ones) so the file is self-contained and ELO is recomputable from it alone. This is why it is large (~16 MB / ~87k rows for LMArena-100k).
- Two `source` vocabularies: `Battle.source` is `llm-judge` | `human` (who decided the outcome); `RatingEntry.source` is `evaluated` | `human` (the model under test vs. every other model on the board).

## Example output

Run: `Qwen2.5-1.5B-Instruct` placed into LMArena-100k, judged by `Qwen3-8B`, 100 battles, 100 bootstraps.

Leaderboard (printed and saved):

```
chatgpt-4o-latest                (4697): 1114.3 ± 4.1
gpt-4o-2024-05-13                (8714): 1088.8 ± 3.2
...
phi-3-mini-4k-instruct           (538) :  890.2 ± 12.0
VLLM/Qwen/Qwen2.5-1.5B-Instruct  (100) <-----: 686.2 ± 47.9
```

`battles.jsonl` (87,331 rows = 100 `llm-judge` + 87,231 `human`):

```json
{"model_a": "VLLM/Qwen/Qwen2.5-1.5B-Instruct", "model_b": "gpt-3.5-turbo-0125", "winner": "model_a", "source": "llm-judge", "question_id": "4c6978df...", "judge_model": "VLLM/Qwen/Qwen3-8B"}
{"model_a": "claude-3-5-sonnet-20240620", "model_b": "gpt-3.5-turbo-0125", "winner": "tie (bothbad)", "source": "human", "question_id": "1a2b3c4d...", "judge_model": null}
```

`elo_ratings.json`:

```json
{
  "arena": "LMArena-100k",
  "model": "VLLM/Qwen/Qwen2.5-1.5B-Instruct",
  "judge_model": "VLLM/Qwen/Qwen3-8B",
  "n_bootstraps": 100,
  "seed": 0,
  "ratings": [
    {"model": "chatgpt-4o-latest", "rating": 1114.3, "ci_low": 1105.9, "ci_high": 1121.8, "n_battles": 4697, "source": "human"},
    {"model": "VLLM/Qwen/Qwen2.5-1.5B-Instruct", "rating": 686.2, "ci_low": 578.8, "ci_high": 780.4, "n_battles": 100, "source": "evaluated"}
  ]
}
```


judgearena/battles.py
```diff
@@ -0,0 +1,111 @@
...
+    models = sorted({m for r in bootstrap_ratings for m in r})
+    entries = []
+    for m in models:
+        vals = np.array([r[m] for r in bootstrap_ratings if m in r], dtype=float)
+        lo, hi = np.percentile(vals, ci)
+        entries.append(
+            RatingEntry(
+                model=m,
+                rating=float(vals.mean()),
+                ci_low=float(lo),
```
ErlisLushtaku, line 113: The console output [here ](https://github.com/OpenEuroLLM/JudgeArena/pull/63/changes#diff-9fee7a3e27c424d4de0a0d4570ed7c81e3f00c1512199ba212b03df4577094f3R439) reports uncertainty as ± np.std(vals), while here we are storing percentile bounds as ci_low / ci_high. Should we align these?

judgearena/battles.py
```diff
@@ -0,0 +1,111 @@
...
+WINNERS = frozenset({"model_a", "model_b", "tie", "tie (bothbad)"})
+
+
+@dataclass(frozen=True)
+class Battle:
+    """One pairwise outcome. ``source`` records its provenance."""
+
+    model_a: str
+    model_b: str
+    winner: str  # one of WINNERS
```
ErlisLushtaku, line 31: Should we persist `pref` / `pref_hard` as well instead of just the winner?

judgearena/estimate_elo_ratings.py
```diff
@@ -419,8 +440,52 @@ def run_judge() -> pd.DataFrame:
     else:
         print("  Not enough data to compute ELO ratings.")
 
+    # Persist artifacts: battles.jsonl is the source of truth (ELO is a pure
+    # function of it); elo_ratings.json + bootstrap_ratings.csv keep the leaderboard.
+    name = (
+        f"{args.arena}-{replace_slash(args.model)}-{replace_slash(args.judge_model)}"
+    )
+    res_folder = Path(args.result_folder) / f"{name}-{datetime.now():%Y%m%d_%H%M%S}"
```
ErlisLushtaku: I think we should use UTC to keep the folder names and metadata consistent

judgearena/estimate_elo_ratings.py
```diff
@@ -419,8 +440,52 @@ def run_judge() -> pd.DataFrame:
...
+        output_dir=res_folder,
+        entrypoint="judgearena.estimate_elo_ratings.main",
+        run=asdict(args),
+        results={
+            **summary,
+            "n_llm_battles": n_llm,
+            "n_human_battles": n_human,
+            "result_folder": str(res_folder),
+        },
+        input_payloads={"instruction_index": question_ids},
```
ErlisLushtaku: question_ids here are arena question_id values, but they're passed under the instruction_index key. The other pipelines (generate_and_evaluate, evaluate) populate with real instruction-dataset row indices. So this run's instruction_indices_sha256 is computed over a different kind of identifier than every other pipeline.

Suggest either renaming the payload key to reflect what it is or passing the genuine df_battles instruction indices under instruction_index (and optionally adding question_id separately) so the hash stays consistent with the other entrypoints.

ErlisLushtaku (commented):
Thanks for the PRs, this is quite small and looks quite good. Just a couple of consistency concerns.

judgearena/battles.py
kargibora: Yes, this is fixed in the new commit. Instead of json, `battles.parquet` carry such information now.

judgearena/estimate_elo_ratings.py
kargibora: I agree, this is resolved.

judgearena/estimate_elo_ratings.py
kargibora: ELO run-metada now uses **question_id**.

kargibora:
Thanks for review @ErlisLushtaku. I have merged with main and updated the code based on yoru comments

judgearena/battles.py
kargibora: I think it makes more sense to use `[ci_low, ci_high]`. I will change the print (std assumes Gaussian dist while ci_low ci_high comptued via percentiles and hence more accurate)

kargibora: [c3178a](https://github.com/OpenEuroLLM/JudgeArena/pull/63/commits/bc3178ae4439d70b5230ba72d745655d7de2199d) fixes this

geoalgo:
Sorry I am just seeing this now. 

> battles.jsonl inlines the human arena battles (not just the judged ones) so the file is self-contained and ELO is recomputable from it alone. This is why it is large (~16 MB / ~87k rows for LMArena-100k).

This seems like a critical point to me. Can we avoid storing this data for every experiment? 
This seems undesirable to me, we are running 1000s of evaluations, we really do not want to store a 16MB which is redundant.
Cant we just load battles from available arenas?

kargibora:
> This seems like a critical point to me. Can we avoid storing this data for every experiment?
> This seems undesirable to me, we are running 1000s of evaluations, we really do not want to store a 16MB which is redundant.
> Cant we just load battles from available arenas?

Hi David, I just removed using jsonl and instead switched to `.parquet`. Of course I can make it optional or discard it completely but parquet is 51x times smaller so the stored information is not 16MB, but something like 0.3 for even 10000 battles. I forgot to change PR description - is this acceptable to you?

geoalgo:
I would rather avoid copying if we can. 
Can we just load the initial tables and join by instruction index?

kargibora:
> I would rather avoid copying if we can. Can we just load the initial tables and join by instruction index?

Yes of course, I will change it

judgearena/battles.py
```diff
@@ -0,0 +1,113 @@
...
+    arena: str
+    model: str
+    judge_model: str
+    n_bootstraps: int
+    seed: int
+    ratings: list[RatingEntry]
+
+    def write(self, path: str | Path) -> None:
+        with Path(path).open("w") as f:
+            json.dump(asdict(self), f, indent=2)
```
ErlisLushtaku: nit: for consistency/NaN-safety with the other writers, we could route through `repro._to_jsonable` and `allow_nan=False`

tests/test_estimate_elo_ratings.py
ErlisLushtaku, line 1: We might need to add a tmp path for tests, or a mock so we don't overwrite the same directory used by the main code if they now save folders.

judgearena/estimate_elo_ratings.py
ErlisLushtaku: I think this is judge-specific now, should we add the judge in the path?

PR #64, opened 2026-06-16
Consolidate CLI arguments into a hierarchical config (refactor 1/N)

### LINKED ISSUE #62: Refactor: one hierarchical run config (CLI + YAML)
## The gist
Before improving the framework more (after Erlis and Alex's PR), I think we should consider a huge (but not so complex) refactoring. My current concern is we are adding too much things and it becomes harder to manage. Reading the entry points at the curent state is quite hard due to lots of argument descriptions, and not enough hierarchy.  I am creating this issue to pinpoint this problem and proposing a solution, which is already done and welcome to any suggestions. 

Configuring a run had grown into a long, flat wall of CLI flags, copy-pasted across the entrypoints. Every new option had to be wired up in several places by hand, nothing was grouped, and there was no way to save or share a run (except from auto-json files). This refactor collapses all of that into a single, structured, typed config that *is* the source of truth — it drives the CLI, an optional YAML file, the `--help` text, and validation, all at once.

The headline: adding an argument used to be a three-place chore; now it's one line.

The branch for the suggested refactoring lies in the [refactor/config-setup](https://github.com/OpenEuroLLM/JudgeArena/tree/refactor/config-setup)

## What you get

- **Everything in one place, grouped by concern.** Run options live in a single config split into `model`, `judge`, `generation`, `elo`, and `run` — so related settings sit together instead of scattered across a flag soup.
- **One source of truth.** The config definition generates the CLI flags, the YAML keys, the `--help` text, and the validation. There's no parallel argument list to keep in sync anymore.
- **Run from a file.** Describe a whole run in YAML and launch it with `--config_path`. Flags still work and override the file; the file overrides the defaults.
- **Self-documenting.** Every field carries a description that shows up in `--help` and in the schema.
- **Reproducible by construction.** Each run drops its fully-resolved config next to the results — point `--config_path` back at it to rerun exactly.

## CLI: before → after

Before — flat flags:

```bash
judgearena --task alpaca-eval \
  --model_A VLLM/Qwen/Qwen2.5-0.5B-Instruct \
  --model_B VLLM/Qwen/Qwen2.5-1.5B-Instruct \
  --judge_model VLLM/Qwen/Qwen2.5-3B-Instruct \
  --n_instructions 10
```

After — grouped flags:

```bash
judgearena --task alpaca-eval \
  --model.name VLLM/Qwen/Qwen2.5-0.5B-Instruct \
  --model.baseline VLLM/Qwen/Qwen2.5-1.5B-Instruct \
  --judge.model VLLM/Qwen/Qwen2.5-3B-Instruct \
  --generation.n_instructions 10
```

…or the same run as a file, with a one-off override on the side:

```bash
judgearena --config_path configs/qwen.yaml
judgearena --config_path configs/qwen.yaml --judge.model VLLM/Qwen/Qwen2.5-7B-Instruct
```

```yaml
# configs/qwen.yaml — only what differs from the defaults needs to be here
task: alpaca-eval
model:
  name: VLLM/Qwen/Qwen2.5-0.5B-Instruct
  baseline: VLLM/Qwen/Qwen2.5-1.5B-Instruct
judge:
  model: VLLM/Qwen/Qwen2.5-3B-Instruct
generation:
  n_instructions: 10
```

## What `--help` looks like now

Because the config defines everything, the help is sectioned by concern, each flag with its own description — not one long flat list (abridged):

```text
$ judgearena --help
usage: judgearena [-h] [--task str] [--model.name {str,null}] [--model.baseline {str,null}]
                  [--judge.model str] [--judge.swap_mode {fixed,both}]
                  [--generation.n_instructions {int,null}] [--elo.n_bootstraps int]
                  [--run.seed int] ...

options:
  -h, --help            show this help message and exit
  --task str            Benchmark to run. Generate+judge: alpaca-eval, arena-hard-v2.0,
                        m-arena-hard-*, mt-bench, fluency-*. ELO: elo-lmarena-*, elo-comparia.

model options:
  Model(s) under evaluation and their generation settings.
  --model.name {str,null}       Model under evaluation, e.g. VLLM/Qwen/Qwen2.5-0.5B-Instruct.
  --model.baseline {str,null}   Opponent for pairwise tasks (defaults to the dataset baseline).
  --model.engine_kwargs dict    Engine kwargs for the backend, e.g. {"tensor_parallel_size": 2}.

judge options:
  The judge model and scoring behaviour.
  --judge.model str               LLM used as the judge, e.g. OpenRouter/deepseek/deepseek-chat-v3.1.
  --judge.swap_mode {fixed,both}  Position-bias handling: single pass vs. both orders.

generation options:
  --generation.n_instructions {int,null}   Number of instructions/battles to evaluate.

elo options:
  Settings specific to elo-rating tasks.
  --elo.n_bootstraps int        Bootstrap resamples for ELO confidence intervals.

run options:
  Seed, output location, caching, and logging.
  --run.seed int                Random seed for reproducibility.
  --run.result_folder str       Where annotations, results, and the resolved config.yaml go.
```

## Adding an argument: before → after

Before, the same option had to be declared in three places that all had to agree: the argparse flag, the dataclass field, and the flag-to-object mapping. Miss one and things drift.

After, you add a single field to the right group:

```python
class JudgeArgs(...):
    top_p: float = 1.0
    """Nucleus sampling top-p for the judge."""
```

That one line gives you the `--judge.top_p` flag, the `judge.top_p` YAML key, the default, the validation, and the `--help` text — for free.

## Breaking changes

- Flags are grouped now: `--model_A` → `--model.name`, `--model_B` → `--model.baseline`, `--judge_model` → `--judge.model`, `--n_bootstraps` → `--elo.n_bootstraps`, and so on. Existing scripts need updating.
- The deprecated `--dataset` / `--arena` aliases are gone — `--task` is the one way in.

## Follow-ups (separate)

- A migration window / temporary aliases for the renamed flags, to land alongside the related breaking-changes issue.
- Optional layering of multiple config files (a shared base plus a small per-run override).

kargibora:
Closes #62.
## Why

The CLI had grown into ~30 flat, loosely-related arguments living in a single `BaseCliArgs` dataclass and duplicated across two entrypoints.
Many were near-duplicate model/judge pairs of the same knob — `--max_out_tokens_models` / `--max_out_tokens_judge`, `--max_model_len` / `--max_judge_model_len`, `--engine_kwargs` / `--judge_engine_kwargs` — with no grouping, so it was hard to tell which flag affected what.
Adding one argument meant touching argparse, the dataclass, and every consumer, and the flags kept drifting apart between the pairwise and ELO paths.

This PR replaces all of that with a single hierarchical `RunConfig` (pydantic-settings), grouped into `model` / `judge` / `generation` / `elo` / `run`.
A new argument is now one field in one place; the CLI flags, the YAML config, validation, and `--help` are all derived from it automatically.

## Before / after

Before:
```bash
judgearena --task elo-lmarena-100k --model_A X --judge_model Y \
  --max_out_tokens_models 2048 --max_out_tokens_judge 8192 \
  --n_bootstraps 100 --soft-elo-temperature 0.5
```

After:
```bash
judgearena --task elo-lmarena-100k --model.name X --judge.model Y \
  --model.max_out_tokens 2048 --judge.max_out_tokens 8192 \
  --elo.n_bootstraps 100 --elo.soft_elo_temperature 0.5

# or drive it from a file (precedence: CLI flags > --config_path YAML > defaults)
judgearena --config_path run.yaml
```

## What's in this PR

- New `judgearena/config.py`: `RunConfig` + grouped sub-models, YAML loading, `--config_path`, and CLI > YAML > defaults precedence.
- Every entrypoint (`generate_and_evaluate`, `estimate_elo_ratings`, `mt_bench`) now reads `cfg.<group>.<field>` instead of flat args.
- Deleted `cli_common.py`; `cli.py` is now a thin dispatcher that builds the config and routes on `--task`.
- Run folders save a round-trippable `config.yaml` (replacing the old `args-*.json`).
- Tests and README updated to the grouped flags.

## Scope

This is the **first in a series** of refactors to make the codebase easier to navigate.
It deliberately keeps the file layout mostly intact and focuses only on consolidating arguments — behaviour is unchanged.
The next steps will **remove and move files** (further module separation and dead-code cleanup).

## Notes

- All tests pass and ruff is clean.
- Most of the diff involves one-line changes (`args.X`->`cfg.X.Y`), and removing old stale builders. 
- We should discuss the namings like `model.name` and `model.baseline`.


kargibora:
Any update on this? It would be great to merge this as my next PR's are built on top of it (so we keep a stable clean config)

judgearena/config.py
```diff
@@ -0,0 +1,312 @@
...
+    """Instruction count and input truncation."""
+
+    elo: EloArgs | None = None
+    """ELO-task settings (only for ``elo-*`` tasks)."""
+
+    run: RunArgs = Field(default_factory=RunArgs)
+    """Run-level settings (seed, output, caching, logging)."""
+
+    @model_validator(mode="after")
+    def _validate(self) -> RunConfig:
```
geoalgo, line 224: Great to have this validator, I wonder if we could also add (in a later PR) a validator for the domains of models, eg. accept things like `Together/meta/llama-70B` but not `Togethers/meta/llama-70B` as we catch this error quite late otherwise.

pyproject.toml
```diff
@@ -38,6 +38,8 @@ dependencies = [
     "pyyaml>=6.0.2",
     "scikit-learn>=1.8.0",
     "tqdm>=4.67.1",
+    "pydantic-settings>=2.0",
+    "pydantic>=2.13.4",
```
geoalgo, line 42: This has to be part of main dependencies because its used in the config right? (if not would be great to have them as dev dependencies)

configs/elo_comparia.yaml
```diff
@@ -0,0 +1,13 @@
+# Example JudgeArena ELO-rating run config.
+# Run with:  judgearena --config_path configs/elo_comparia.yaml
+# ELO tasks rate a single model (model.name) against arena opponents; the
+# arena is derived from the elo-* task, so elo.arena does not need to be set.
+task: elo-comparia
+model:
+  name: Together/meta-llama/Llama-3.3-70B-Instruct-Turbo
+judge:
+  model: OpenRouter/deepseek/deepseek-chat-v3.1
```
geoalgo: Can we rather set gemma4 which is our default judge?

configs/elo_comparia.yaml
```diff
@@ -0,0 +1,13 @@
...
+generation:
+  n_instructions: 200
+elo:
```
geoalgo, line 12: aren't we missing languages?

configs/alpaca_eval.yaml
```diff
@@ -0,0 +1,12 @@
+# Example JudgeArena run config (pairwise generate + judge).
+# Run with:  judgearena --config_path configs/alpaca_eval.yaml
+# Only the fields you want to change need to be present; everything else uses
+# the built-in defaults. CLI flags (e.g. --judge.model X) override this file.
+task: alpaca-eval
+model:
+  name: gpt4_1106_preview
+  baseline: VLLM/utter-project/EuroLLM-9B
+judge:
+  model: OpenRouter/deepseek/deepseek-chat-v3.1
```
geoalgo: Can we rather set gemma4 which is our default judge?

geoalgo (commented):
LGTM, mostly have small comments.

judgearena/config.py
kargibora: Yes, that would be an easy step with wrappers. But does this require a set of pre-defined domains?

pyproject.toml
kargibora: Yes, pydantic is a commonly used dependency in almost any new frameworks. It should stay if we want to use for example automatic helper docstrings from the class

configs/alpaca_eval.yaml
kargibora: These were more or less some example configs but of course, I will swap gemma4

configs/elo_comparia.yaml
kargibora: Yes of course

configs/elo_comparia.yaml
kargibora: I will add it as well, missed it as I have tried to make the example yamls very simple (if default values are setting the languages than it would still work)

geoalgo: Perhaps we can set the default to just english and indicate as a comment how to include other languages.

pyproject.toml
geoalgo: One point is just that it may easily introduce conflict (its a typical situation to have conflict in your deps about pydantic), could we move it to dev dependencies in a later PR?

judgearena/config.py
geoalgo: Fair point it will require having a list (I thought we were doing it but we are not). I think it will be valuable as users (and myself :-)) often pass a model without the domain which fails quite late. Should be done in a later PR, perhaps we can just leave a todo comment for now.

pyproject.toml
kargibora: VLMM and OpenAI API, from what I know, already depends on pydantic. I need to re-test it whether it works without it us specifying a version as well (in a later PR). I agree that `pydantic` version can conflict with framework if we are using an old library that requires `pydantic<2.0.0`.

kargibora:
If everything looks good to you, I will be merging this @geoalgo . Thanks for the review!

configs/elo_comparia.yaml
```diff
@@ -0,0 +1,16 @@
+# Example JudgeArena ELO-rating run config.
+# Run with:  judgearena --config_path configs/elo_comparia.yaml
+# ELO tasks rate a single model (model.name) against arena opponents; the
+# arena is derived from the elo-* task, so elo.arena does not need to be set.
+task: elo-comparia
+model:
+  name: Together/meta-llama/Llama-3.3-70B-Instruct-Turbo
+judge:
+  model: OpenRouter/google/gemma-4-26b-a4b-it
```
PriorDavid: We use [google/gemma-4-31B](https://huggingface.co/google/gemma-4-31B) by default in the paper could we use this one here too? (also from other yaml)

PriorDavid (approved):
LGTM, can we just change to [google/gemma-4-31B](https://huggingface.co/google/gemma-4-31B) which we use in the paper?

geoalgo (approved):

kargibora:
> LGTM, can we just change to [google/gemma-4-31B](https://huggingface.co/google/gemma-4-31B) which we use in the paper?

I have updated it with the latest commit (should I make it VLLM or OpenRouter)?

geoalgo:
perhaps vllm is slightly better but either way is fine

PR #65, opened 2026-06-19
Split `utils` into a package + extract a models layer (refactor 2/N)

kargibora:
`utils.py` had grown into a ~500-line grab-bag: model wrappers and inference, file/HF I/O, text helpers, preference/eval statistics, and assorted constants all in one module. Importing one helper pulled in the whole thing, the boundaries between "how we call models" and "how we score them" were invisible, and related constants were scattered across modules.

This PR splits that single module along its real responsibilities, so each file has one job and call sites import from an obvious place. Behaviour is unchanged — this is a move/rename refactor.

## What's in this PR

- **`utils.py` → `judgearena/utils/` package**, split by responsibility:
  - `utils/io.py` — file / Hugging Face dataset I/O.
  - `utils/text.py` — text helpers (truncation, thinking-tag stripping, …).
  - `utils/eval.py` — preference statistics / eval helpers.
  - `utils/__init__.py` — re-exports so existing `from judgearena.utils import …` keeps working.
- **New `judgearena/models.py`** — the model-wrapper / inference layer extracted out of utils.
- **New `judgearena/constants.py`** — centralizes constants (e.g. the ELO task→arena map and task prefixes) that were previously scattered.
- **Removed `eval_utils.py` and `tasks.py`** — folded into the package / constants.
- All consumers updated (`evaluate`, `generate`, `generate_and_evaluate`, `estimate_elo_ratings`, `mt_bench/*`, `config`) plus the tests.

## Scope

Second in the refactor series (follows the config consolidation, #64). Where the config PR deliberately kept the file layout intact, this one **removes and moves files** for real module separation. Still behaviour-preserving — no functional changes.

## Notes

- All tests pass and ruff is clean.
- Base this against `refactor/config-on-main`; the branch already merges it, so the diff is just the utils reorganization
- This is, similar to the previous refactoring PR, a simple moving of the existing functionailty. Thus it does not introduce any new class or scheme, just organizes it so in the future PR's like `leaderboard`, we can put what necessary to the more fitting scripts.


judgearena/utils/io.py
```diff
@@ -0,0 +1,187 @@
...
+    if raw is None or not raw.strip():
+        return None
+    try:
+        return int(raw)
+    except ValueError:
+        logger.warning("Ignoring malformed %s=%r; expected an integer.", env_var, raw)
+        return None
+
+
+def download_all():
```
ErlisLushtaku, line 72: This drops the `revision=hf_revision("geoalgo/multilingual-contexts-to-be-completed")` pin that's on main. Is that intentional?

judgearena/models.py
```diff
@@ -110,30 +102,6 @@ def _resolve_chat_template_kwargs(
     return chat_template_kwargs or None
 
 
-def set_langchain_cache():
```
ErlisLushtaku, line 114: This has not been readded, just removed. Was it not used?

kargibora: Current only use case is under `scripts/fluency`,  I think we should better define it there if we are not going to use it.

judgearena/utils/io.py
kargibora: No that is a bug, thanks for catching it

judgearena/models.py
ErlisLushtaku, line 1: The imports of the exported members from this file should be updated, e.g. the scripts such as [generate_fluency](https://github.com/OpenEuroLLM/JudgeArena/blob/2a0946088c865a172a2bdde2e1f256c613dc3466/scripts/fluency/generate_fluency.py#L6)

judgearena/estimate_elo_ratings.py
```diff
@@ -19,8 +19,12 @@
 )
 from judgearena.generate import generate_instructions
 from judgearena.log import get_logger
+from judgearena.models import make_model
 from judgearena.repro import _to_jsonable
-from judgearena.utils import cache_function_dataframe, compute_pref_summary, make_model
+from judgearena.utils import cache_function_dataframe, compute_pref_summary
+
+if TYPE_CHECKING:
+    from judgearena.config import RunConfig
```
ErlisLushtaku: Duplicate of the code below

judgearena/models.py
ErlisLushtaku: Okay, makes sense

judgearena/models.py
kargibora: Fixed it [564f4a9](https://github.com/OpenEuroLLM/JudgeArena/pull/65/commits/564f4a99bc3b2cdd7af08870b93c3cf4215b001d)

judgearena/estimate_elo_ratings.py
kargibora: Fixed it [564f4a9](https://github.com/OpenEuroLLM/JudgeArena/pull/65/commits/564f4a99bc3b2cdd7af08870b93c3cf4215b001d)

judgearena/utils/io.py
kargibora: Fixed it [564f4a9](https://github.com/OpenEuroLLM/JudgeArena/pull/65/commits/564f4a99bc3b2cdd7af08870b93c3cf4215b001d)

ErlisLushtaku (approved):
LGTM, only CI needs to go green

kargibora:
Merging this.

PR #66, opened 2026-06-19
Typed result reporting: `PrefSummary` / `Report` / `BattleReport` / `EloReport` (refactor 3/N)

kargibora:
The three pipelines (arena, MT-bench, ELO) each hand-assembled an untyped `results` dict and reported it inconsistently. There were **two diverged `print_results` functions** with the same name but different behaviour — one handled `per_category`/`per_turn`, the other `swap_mode`/`result_folder`/unparseable-warnings — so it was easy to call the wrong one. Untyped access (`results['task']`, `results.get(...)`) meant typos surfaced only at runtime, ELO's reporting was bespoke `print(...)`s tangled through `main()`, and "how a result is persisted" was split between each pipeline's caller and a one-off `write_elo_result`.

This PR replaces the ad-hoc dicts and duplicated printers with typed report objects that own their rendering, serialization, and saving, sharing a single statistics core.

## What's in this PR

- **`PrefSummary`** — a dataclass for win/loss/tie stats; `compute_pref_summary` now returns it, so consumers use attribute access instead of dict keys.
- **`Report` ABC — the full persistence contract.** Subclasses implement `render()` (human output) and `_payload()` (their fields); the base provides `to_dict()` (a versioned envelope — `schema_version` + `report_type` + the payload) and a uniform `save(path)`. So every report saves the same way and every payload is self-describing and migratable.
- **`BattleReport`** — one report for both the arena and MT-bench pipelines, unifying the two `print_results` into a single `render()`. Its `metadata` is now **nested** under a `metadata` key rather than spread into the top level, so adding new saved fields is safe and a metadata key can't silently clobber a core one. Both old `print_results` are deleted; `generate_and_evaluate.py` / `mt_bench_utils.py` build a `BattleReport` and persist via `report.save(...)`, and `utils/__init__` drops the `print_results` re-export.
- **`EloReport`** — replaces `write_elo_result`; `estimate_elo_ratings.main()` is reordered to compute everything → build the report → `render()` / `save()`.
- Tests: `tests/test_eval_reports.py` covers the summary, both `BattleReport` shapes, the versioned envelope, `EloReport`, and a `save()` round-trip.

## Scope

Third in the refactor series . The run pipelines and `main()` return contracts are unchanged; the **saved results JSON gains a versioned schema (v1)** — payloads now carry `schema_version` + `report_type`, and `BattleReport` metadata is nested under `metadata`. No in-repo code reads those fields back, so this is self-contained; external analysis that reads the old flat metadata fields needs the nested path.

## Notes

- All tracked tests pass and ruff is clean.
- Not so important (unlike the #64), however will keep the things on shape for printing/saving results thus following PR's are better handled
- Most of the LoC is coming from the test (+200), others are refactoring. We can also reduce the test coverage as it may not be super important to chekc static types like this with a test (or merge some tests together)


judgearena/estimate_elo_ratings.py
```diff
@@ -182,23 +184,108 @@ def _slugify(value: str) -> str:
...
+            "llm_judged_battles": self.llm_judged_battles,
+            "human_anchor_battles": self.human_anchor_battles,
+            "sampling_metadata": self.sampling_metadata,
+            "elo_mean": self.elo_mean,
+            "elo_std": self.elo_std,
+            "elo_num_bootstraps": self.elo_num_bootstraps,
+            "source_battle_counts": self.source_battle_counts,
+        }
+
+    def save(self, result_folder: str | Path) -> Path:
```
ErlisLushtaku: This override breaks the uniform `save()` contract the PR introduces: the base `save(path)` in [eval.py](https://github.com/OpenEuroLLM/JudgeArena/pull/66/changes#diff-74b92b89d6bc74c033df107365fc02a9ab3de893bedb3416be3556d62aa4fdc8R69) which writes the envelope to a file path, but here the argument is a folder, the output path is derived internally, and the envelope is nested under a "summary" key plus bootstrap_ratings. So ELO files and BattleReport files end up with different top-level shapes, and report.save(x) means different things per subtype. I would suggest we make it consistent.

judgearena/estimate_elo_ratings.py
```diff
@@ -182,23 +184,108 @@ def _slugify(value: str) -> str:
...
+    summary: PrefSummary
+    num_battles: int
+    llm_judged_battles: int
+    human_anchor_battles: int
+    elo_mean: float
+    elo_std: float
+    elo_num_bootstraps: int
+    mae_vs_human: float
+    method: str
+    calibrated_temperature: float
```
ErlisLushtaku: nit: is this field used anywhere?

judgearena/estimate_elo_ratings.py
```diff
@@ -182,23 +184,108 @@ def _slugify(value: str) -> str:
...
-        "bootstrap_ratings": bootstrap_ratings,
-    }
-    path.write_text(json.dumps(_to_jsonable(payload), indent=2) + "\n")
-    return path
+@dataclass
+class EloReport(Report):
+    """Bradley-Terry / Soft-ELO ratings for one focal model against an arena."""
+
+    arena: str
+    model_a: str
```
ErlisLushtaku: nit: do we need both model_a and [model_name](https://github.com/OpenEuroLLM/JudgeArena/pull/66/changes#diff-9fee7a3e27c424d4de0a0d4570ed7c81e3f00c1512199ba212b03df4577094f3R205)? Is there risk of these two drifting if we don't use one of them consistently?

ErlisLushtaku (approved):
LGTM, just a couple of small comments in case you think they are valid

kargibora:
Hi @ErlisLushtaku, sorry to request this from you again but I have resolved the problems as you mentioned and changed some stuff by using pydantic - so we dont have to serialize everything again and it serialized and validated automatically. Can you take a look?

ErlisLushtaku:
> Hi @ErlisLushtaku, sorry to request this from you again but I have resolved the problems as you mentioned and changed some stuff by using pydantic - so we dont have to serialize everything again and it serialized and validated automatically. Can you take a look?

LGTM

ErlisLushtaku (approved):

PR #58, opened 2026-06-01
5.5. random 1k battles elo sampling

ErlisLushtaku:
## Summary
- Add Elo sampling controls and CLI wiring.

PR #69, opened 2026-06-30
Use strip-thinking and thinking budget in LMArena Elo path

ErlisLushtaku:
## Summary
- The LMArena Elo entrypoint (`estimate_elo_ratings.py`) accepted `--strip_thinking_before_judging` and `--battle_thinking_token_budget` via `CliEloArgs` but **silently ignored both**: generation built engine kwargs without the thinking-token sub-budget, and the judge call omitted the strip flag. This made Elo inconsistent with the benchmark path and produced ranking anomalies (e.g. Olmo-3-7B-Think below Olmo-3-7B-Instruct).
- Generation now reuses `_build_generation_engine_kwargs` to inject the thinking-token budget for VLLM thinking models (capped by `max_out_tokens_models`).
- The judge call forwards `strip_thinking_before_judging`.
- The judge cache suffix gains a `_stripthinking` marker (only when enabled) so prior non-stripped caches invalidate while existing hashes are preserved.
- `_is_retryable_error` now retries transient `json.JSONDecodeError` raised when an OpenRouter gateway returns non-JSON (HTML) error bodies.
- `--provide_explanation` / `--strip_thinking_before_judging` accept explicit `true/false` via `str2bool`, keeping bare-flag behavior.

## Why stacked here
This depends on the Elo entrypoint introduced in `pr32-split-v3/05.5-elo-sampling`, which is not yet in `main`, so the PR targets that branch.

## Test plan
- [x] `tests/test_estimate_elo_ratings.py` — strip flag forwarded, thinking-budget injected/capped, absent for non-thinking models (15 passed)
- [x] `tests/test_cli.py` — `str2bool` flag parsing (30 passed)
- [x] `tests/test_utils.py` — `_is_retryable_error` retries JSONDecodeError / transient HTTP, not auth (passed)
- [x] `ruff check` + `ruff-format` clean

Includes-AI-Code: true

### DIFF
--- a/judgearena/cli_common.py
+++ b/judgearena/cli_common.py
@@ -13,6 +13,26 @@ from dataclasses import dataclass, field
 
 from judgearena.prompts.registry import JUDGE_PROMPT_PRESETS
 
+_TRUE_TOKENS = {"true", "t", "yes", "y", "1"}
+_FALSE_TOKENS = {"false", "f", "no", "n", "0"}
+
+
+def str2bool(value: str | bool) -> bool:
+    """Parse a CLI boolean supplied either as a bare flag or an explicit value.
+
+    Slurmpilot serialises ``python_args`` dicts as ``--key=value``, so boolean
+    flags reach the CLI as ``--flag=True`` / ``--flag=False`` instead of bare
+    switches. Accept both forms so launcher-driven and interactive use agree.
+    """
+    if isinstance(value, bool):
+        return value
+    normalized = value.strip().lower()
+    if normalized in _TRUE_TOKENS:
+        return True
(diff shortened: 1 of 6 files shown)


kargibora:
Hi Erlis, is there a way for you to change the target to main instead? Because lots of refactorings happened since and possibly these changes might be targeted for a stale branch

ErlisLushtaku:
Superseded by #70, which targets `main` directly. PR #58 (`pr32-split-v3/05.5-elo-sampling`) is already merged into main, so this stacked PR is no longer needed. The fix is re-implemented in #70 against the current post-refactor `main` structure.

PR #70, opened 2026-06-30
Honor strip-thinking and thinking budget in LMArena Elo path

ErlisLushtaku:
## Summary
The LMArena Elo entrypoint (`estimate_elo_ratings.py`) exposed `strip_thinking_before_judging` and `battle_thinking_token_budget` (via `cfg.judge`) but **silently ignored both**: generation called `cfg.model.evaluated_generation_kwargs()` directly (no thinking-token sub-budget) and the judge call omitted the strip flag. This made Elo inconsistent with the benchmark path and produced ranking anomalies (e.g. Olmo-3-7B-Think below Olmo-3-7B-Instruct).

- Generation reuses `_build_generation_kwargs(cfg, ..., role="A")` so VLLM thinking models receive the thinking-token budget (capped by `max_out_tokens`), matching `generate_and_evaluate` and `mt_bench`.
- The judge call forwards `strip_thinking_before_judging` (already supported by `judge_and_parse_prefs`).
- The judge cache suffix gains a `_stripthinking` marker (only when enabled) so prior non-stripped caches invalidate while existing hashes are preserved.
- `models._is_retryable_error` now retries transient `json.JSONDecodeError` raised when an OpenRouter gateway returns non-JSON (HTML) error bodies.

## Notes
- Supersedes #69, which targeted the now-merged `pr32-split-v3/05.5-elo-sampling` (PR #58). This PR is rebased onto current `main` and adapted to the post-refactor structure (`cfg`/pydantic config, `models.py`, `utils/` package). The earlier `str2bool`/`cli_common` changes are obsolete because pydantic already coerces `--flag=true/false`.

## Test plan
- [x] `tests/test_estimate_elo_ratings.py` — strip flag forwarded (on/off), thinking-budget injected/capped/absent (passed)
- [x] `tests/test_chat_vllm.py` — `_is_retryable_error` retries JSONDecodeError / transient HTTP, not auth (passed)
- [x] 31 passed across both files; `ruff check` + `ruff-format` clean; pre-commit hooks pass

Includes-AI-Code: true

### DIFF
--- a/judgearena/estimate_elo_ratings.py
+++ b/judgearena/estimate_elo_ratings.py
@@ -18,6 +18,7 @@ from judgearena.evaluate import (
     resolve_run_judge_prompt,
 )
 from judgearena.generate import generate_instructions
+from judgearena.generate_and_evaluate import _build_generation_kwargs
 from judgearena.log import get_logger
 from judgearena.models import build_default_judge_model_kwargs, make_model
 from judgearena.repro import _to_jsonable
@@ -301,7 +302,11 @@ def main(cfg: "RunConfig") -> dict:
     # Step 2: Generate completions for the model under evaluation
     logger.info("Step 2: Generating completions with %s", cfg.model.name)
 
-    extra_kwargs = cfg.model.evaluated_generation_kwargs()
+    # Mirror the benchmark generation path so Elo battles honor the
+    # thinking-token sub-budget for thinking models (the Elo entrypoint
+    # previously called evaluated_generation_kwargs() directly and silently
+    # dropped battle_thinking_token_budget).
+    extra_kwargs = _build_generation_kwargs(cfg, cfg.model.name, role="A")
     use_tqdm = False
     gen_fun = partial(
         generate_instructions,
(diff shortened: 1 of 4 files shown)


judgearena/estimate_elo_ratings.py
```diff
@@ -301,7 +302,11 @@ def main(cfg: "RunConfig") -> dict:
     # Step 2: Generate completions for the model under evaluation
     logger.info("Step 2: Generating completions with %s", cfg.model.name)
 
-    extra_kwargs = cfg.model.evaluated_generation_kwargs()
+    # Mirror the benchmark generation path so Elo battles honor the
+    # thinking-token sub-budget for thinking models (the Elo entrypoint
+    # previously called evaluated_generation_kwargs() directly and silently
+    # dropped battle_thinking_token_budget).
+    extra_kwargs = _build_generation_kwargs(cfg, cfg.model.name, role="A")
```
kargibora, line 309: Thanks for catching this.

kargibora (commented):
LGTM.

PR #71, opened 2026-07-08
Cache

geoalgo:


judgearena/store_sqlite.py
```diff
@@ -0,0 +1,280 @@
+"""SQLite-backed completion and judgement store.
```
kargibora, line 1: I have several concerns for this:


**(1)**  We are not using model/judge/prompt configurations in the index. Maybe this is resolved later but any change to system prompt, in my opinion, should be reflected in a key (like some unique hash maybe)?

**(2)**      `instruction_index INTEGER PRIMARY KEY` Works just fine for datasets where `instruction_index` is unique hash key, however if it is an integer for row index at some point, (for example ELO pipeline we have), then different filterings yielding same dataset idx can query the cache even if the language for example is entirely wrong.

**(3)** Related to **(1)**, ` UNIQUE(instruction_index, model_A, model_B)` does not include the content that is judged. If for example, we use a different temperature for `model_A` we will still get the old judgement even if the generated output is different.

judgearena/store_sqlite.py
```diff
@@ -0,0 +1,280 @@
...
+        return len(rows)
+
+    def query(self, indices: list[int] | None = None) -> pd.DataFrame:
+        """Return completions, optionally filtered to specific instruction_indices."""
+        conn = self._connect()
+        if indices is None:
+            return pd.read_sql(
+                "SELECT * FROM completions ORDER BY instruction_index", conn
+            )
+        placeholders = ",".join("?" * len(indices))
```
kargibora, line 100: This will fail if `len(indices)` is too large. 
> The Limit: 32,766 parameters

If we want to get ~40.000 row this will fail. Need some chunk up loading mechanism (would allow parallization as well)

judgearena/generate.py
```diff
@@ -12,40 +25,64 @@ def generate_instructions(
     max_tokens: int | None = 32768,
     use_tqdm: bool = True,
     system_prompt: str | None = None,
+    completion_store: SQLiteCompletionStore | None = None,
+    pushed_by: str = "judgearena",
     **engine_kwargs,
 ) -> pd.DataFrame:
+    # Filter to instructions not already in the shared store
+    if completion_store is not None:
+        all_indices = instructions.index.tolist()
```
kargibora, line 34: We should be sure about `instructions.index` is never changes under the filtering steps (category/language slices etc).

judgearena/estimate_elo_ratings.py
```diff
@@ -11,389 +11,234 @@
...
+                "pt",
+                "ru",
+                "vi",
+                "no",
+                "fr",
+                "pl",
+                "sr",
+                "ar",
+            ]
+        print(f"filtering with languages: {args.languages}")
```
kargibora, line 228: Maybe lets use `logging.debug` as we are using it before so we keep track of the logs easily (also the users)

judgearena/estimate_elo_ratings.py
```diff
@@ -496,72 +338,122 @@ def replace_slash(s: str) -> str:
...
-            row_use_model_a = use_model_a_as_opponent
-            row_our_pos_a = our_model_is_position_a
-            row_opponents = list(opponent_models)
+        # Determine which indices still need judging
+        all_indices = list(range(n))
+        if judgement_store is not None:
+            pairs = list(
+                zip(all_indices, model_A_per_row, model_B_per_row, strict=True)
+            )
+            cached_df = judgement_store.query(model=args.model)
```
kargibora, line 364: `cached_df` contains every cached row for the evaluated model, but it is not restricted to the keys requested by the current run before being concatenated with new_df. This can include stale or out-of-range rows when sample size, language filters, or seed change.

judgearena/estimate_elo_ratings.py
```diff
@@ -496,72 +338,122 @@ def replace_slash(s: str) -> str:
...
+        new_df = pd.DataFrame()
+        if missing_idx:
+            judge_chat_model = make_model(
+                model=args.judge_model,
+                max_tokens=args.max_out_tokens_judge,
+                **judge_extra_kwargs,
             )
-            row_our_pos_a = np.concatenate(
-                [our_model_is_position_a, our_model_is_position_a]
+            annotations, _, prefs = judge_and_parse_prefs(
```
kargibora, line 394: We give `args.swap_mode`, if its both, then both annotation and reversed annotation is returned. However from what I know prefs are `2m` 

We run this function in `judge_and_parse_prefs`:
```py
def combine_swapped_prefs(prefs_ab: pd.Series, prefs_ba: pd.Series) -> pd.Series:
    """Combine swap_mode='both' prefs into one P(B wins) series: [pref_AB, 1 - pref_BA].

    ``prefs_ab`` are P(B wins) from the AB ordering; ``prefs_ba`` are P(B wins)
    from the swapped BA ordering, so ``1 - prefs_ba`` re-orients them to the AB
    frame before stacking.
    """
    return pd.concat(
        [prefs_ab.reset_index(drop=True), 1 - prefs_ba.reset_index(drop=True)]
    ).reset_index(drop=True)

prefs = combine_swapped_prefs(prefs.apply(_none_to_nan), prefs_reversed)
```

judgearena/browse_cache.py
```diff
@@ -0,0 +1,900 @@
...
+            "---",
+            "### Judge input",
+            "*(not stored — judge_input was not saved during migration)*",
+        ]
+
+    return "\n\n".join(parts)
+
+
+try:
+    from textual import work
```
kargibora, line 654: Should we add this to `pyproject.toml`?

PR #75, opened 2026-07-17
Remove stale evaluation helpers

kargibora:
## Summary

This PR removes two unused legacy helpers from `judgearena/evaluate.py`:

- `evaluate_completions()`
- `load_judge_system_and_user_prompt()`

It also removes their unused imports and updates the obsolete prompt-resolution reference in the `annotate_battles()` documentation.

## Why

`evaluate_completions()` is not used by the CLI, internal pipelines, tests, or documentation. Its responsibilities are already handled by the config-driven `generate_and_evaluate` pipeline, which supports the current task, model, judge, prompt, caching, and output configuration.

`load_judge_system_and_user_prompt()` also has no callers. Prompt resolution is handled by `resolve_judge_prompts()` and the prompt registry, including presets, custom prompt files, task defaults, and explanation settings.

Keeping these alternative paths increases the maintenance surface and requires new pipeline features to support an undocumented execution flow that is no longer used.

## Compatibility

The JudgeArena CLI and active evaluation pipelines are unchanged. Low-level APIs such as `annotate_battles()` and `judge_and_parse_prefs()` remain available.

The removed helpers were not exported from the package root or documented as public APIs. However, code importing them directly from `judgearena.evaluate` will need to migrate to the config-driven pipeline or the current low-level APIs.

### DIFF
--- a/judgearena/evaluate.py
+++ b/judgearena/evaluate.py
@@ -1,20 +1,11 @@
-import json
 import re
 from dataclasses import dataclass
-from datetime import UTC, datetime
-from pathlib import Path
 
 import numpy as np
 import pandas as pd
-from langchain_core.language_models.llms import LLM
 from langchain_core.prompts import ChatPromptTemplate
 from scipy.optimize import minimize_scalar
 
-from judgearena.instruction_dataset import load_instructions
-from judgearena.instruction_dataset.arena_hard import (
-    download_arena_hard,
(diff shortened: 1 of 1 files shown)


ErlisLushtaku (approved):

kargibora:
I am pushing this. We can additionally implement more cleaner API function if we want to expose such functions to people. This was quite outdated

PR #76, opened 2026-07-18
Add meta evaluation task

ErlisLushtaku:
## Summary

- Add `--task meta-eval` for evaluating judge models against human-labeled arena battles.
- Support deterministic sampling, language filtering, four named prompt modes, and optional swapped-order judging.
- Report agreement, ranking, ELO, bootstrap uncertainty, ELO-gap, token, and cost metrics.
- Save auditable per-pass annotations, summaries, results, and run metadata.
- Package prompt resources and integrate meta-eval with the hierarchical `RunConfig` CLI.

## Usage

```bash
judgearena \
  --task meta-eval \
  --meta_eval.reference_arena LMArena-140k \
  --meta_eval.prompt_mode standard \
  --meta_eval.languages '["en", "es"]' \
  --meta_eval.top_models 20 \
  --meta_eval.n_bootstraps 1000 \
  --run.seed 0
```

## Outputs

Runs produce:

- `annotations.parquet`
- `results.json`
- `summary.csv`
- `args.json`
- `run-metadata.v1.json`

With `--judge.swap_mode both`, annotations contain one row per judge pass with normalized model ordering and an `orientation` column.

## Cache note

Meta-eval temporarily retains the existing per-battle SQLite WAL cache. It should be used from a single host and not concurrently across NFS-mounted compute nodes. Cache unification remains deferred to the dedicated caching work.

### DIFF
--- a/judgearena/meta_eval/annotate.py
+++ b/judgearena/meta_eval/annotate.py
@@ -1,0 +1,273 @@
+"""Run LLM judge annotations for meta-evaluation battles."""
+
+from __future__ import annotations
+
+import pandas as pd
+
+from judgearena.arenas_utils import extract_turn_text
+from judgearena.evaluate import JudgeAnnotation, annotate_battles
+from judgearena.meta_eval.cache import (
+    AnnotationCache,
+    AnnotationEntry,
+    AnnotationKey,
+)
+from judgearena.meta_eval.cli_args import CliMetaEvalArgs
+from judgearena.meta_eval.cost import (
+    estimate_annotation_cost_usd,
+    estimate_token_count,
+)
+from judgearena.meta_eval.parsers import add_parsed_columns, invert_winner
+
+
+def _battle_texts(df_batch: pd.DataFrame) -> tuple[list[str], list[str], list[str]]:
+    instructions = [extract_turn_text(conv[0]) for conv in df_batch["conversation_a"]]
+    completions_a = [
+        extract_turn_text(conv[1]) if len(conv) > 1 else ""
+        for conv in df_batch["conversation_a"]
+    ]
+    completions_b = [
+        extract_turn_text(conv[1]) if len(conv) > 1 else ""
+        for conv in df_batch["conversation_b"]
+    ]
+    return instructions, completions_a, completions_b
+
+
+def _swap_batch(df_batch: pd.DataFrame) -> pd.DataFrame:
+    swapped = df_batch.copy()
+    swapped["conversation_a"] = df_batch["conversation_b"]
+    swapped["conversation_b"] = df_batch["conversation_a"]
+    swapped["model_a"] = df_batch["model_b"]
+    swapped["model_b"] = df_batch["model_a"]
+    return swapped
+
+
+def _annotations_to_frame(
+    df_batch: pd.DataFrame,
+    annotations,
+    *,
+    prompt_mode: str,
+    judge_model: str,
+) -> pd.DataFrame:
+    rows = []
+    for ann, (_, battle) in zip(annotations, df_batch.iterrows(), strict=True):
+        judge_input = ann.judge_input or ""
+        cost_usd, cost_source = estimate_annotation_cost_usd(
+            judge_input=judge_input,
+            judge_completion=ann.judge_completion,
+            judge_model=judge_model,
+        )
+        rows.append(
+            {
+                "question_id": battle["question_id"],
+                "model_a": battle["model_a"],
+                "model_b": battle["model_b"],
+                "winner": battle["winner"],
+                "lang": battle["lang"],
+                "benchmark": battle["benchmark"],
+                "instruction": ann.instruction,
+                "completion_a": ann.completion_A,
+                "completion_b": ann.completion_B,
+                "judge_input": ann.judge_input,
+                "judge_completion": ann.judge_completion,
+                "estimated_input_tokens": estimate_token_count(judge_input),
+                "estimated_output_tokens": estimate_token_count(ann.judge_completion),
+                "cost_usd": cost_usd,
+                "cost_source": cost_source,
+            }
+        )
+    return add_parsed_columns(pd.DataFrame(rows), prompt_mode)
+
+
+def _judge_cache_name(args: CliMetaEvalArgs) -> str:
+    if args.prompt_mode == "standard":
+        return args.judge_model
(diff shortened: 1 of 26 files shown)


judgearena/meta_eval/parsers.py
```diff
@@ -0,0 +1,147 @@
...
+    temperature: float = META_EVAL_PAIRSCORE_TEMPERATURE,
+) -> PairScore:
+    parser = PairScore()
+    parser.temperature = temperature
+    return parser
+
+
+def parse_pairscore_pref(judge_completion: str, *, temperature: float) -> float:
+    score = pair_score_parser(temperature).parse_model_raw(judge_completion)
+    if score is None or np.isnan(score):
```
kargibora, line 41: Why we return `tie=0.5`  for parsing errors? We have to make sure this does not change the Cohen's Kappa or any accuracy

judgearena/meta_eval/metrics.py
```diff
@@ -0,0 +1,439 @@
...
+                model_mask_top = (df_battles["model_a"] == model) | (
+                    df_battles["model_b"] == model
+                )
+                other_human = df_battles[~model_mask_top].copy()
+                sample = df_ann[model_mask_ann].sample(
+                    n=num_battles,
+                    replace=False,
+                    random_state=int(rng.integers(0, 2**32 - 1)),
+                )
+                if exclude_ties:
```
kargibora, line 367: Sampling `num_battles` row then excluding tie means that our battles will have `n <= num_battles`

judgearena/meta_eval/runner.py
```diff
@@ -0,0 +1,259 @@
...
+        attach_file_handler(make_run_log_path(res_folder))
+
+    with open(res_folder / "args.json", "w", encoding="utf-8") as handle:
+        json.dump(asdict(args), handle, indent=2)
+
+    prompt_spec = resolve_prompt_mode(
+        args.prompt_mode,
+        provide_explanation=args.provide_explanation,
+    )
+    judge_chat_model = make_model(
```
kargibora, line 191: Should we call `build_default_judge_model_kwargs` before? Or this is already enough

judgearena/meta_eval/annotate.py
```diff
@@ -0,0 +1,273 @@
...
+def annotate_sample(
+    df_sample: pd.DataFrame,
+    args: CliMetaEvalArgs,
+    *,
+    judge_chat_model,
+    prompt_spec,
+    annotation_cache: AnnotationCache | None = None,
+) -> pd.DataFrame:
+    n_total = len(df_sample)
+    n_batches = (n_total + args.batch_size - 1) // args.batch_size
```
kargibora, line 227: Not so problematic but we never check `args.batch_size > 0`. It can raise division by zero by mistake.

kargibora:
Apart from these I feel like there are some duplications:

**(1)**
`compute_soft_bradley_terry()` reimplements most of fit_bradley_terry().

  The existing function already supports continuous preferences:

```py
  fit_bradley_terry(df, pref_col="pref_llm")
```

**(2)**

The PR introduces:
```py
  PromptModeSpec
  resolve_prompt_mode()
  _read_prompt()
```
  JudgeArena already has:

```py
  JudgePromptPreset
  ResolvedJudgePrompt
  resolve_judge_prompt()
```

Are these exchangable or we are required to implement this specifically?

judgearena/meta_eval/metrics.py
ErlisLushtaku: `num_battles` is the annotation budget for this analysis. The reference samples that many annotations and then drops tie predictions [here](https://github.com/geoalgo/multilingual-llm-judge/blob/868ff7bba73fc56db7c23cb67d1601531a2dd533/scripts/benchmark_judge.py#L221-L227), so the Bradley-Terry fit can use fewer decisive rows. I kept this behavior and clarified it [here](https://github.com/OpenEuroLLM/JudgeArena/commit/03f8e5a54359e91e509b118c64c3c68f4c8877c0).

judgearena/meta_eval/annotate.py
ErlisLushtaku: Good catch, added the validation [here](https://github.com/OpenEuroLLM/JudgeArena/commit/03f8e5a54359e91e509b118c64c3c68f4c8877c0).

judgearena/meta_eval/parsers.py
ErlisLushtaku: Yes, malformed outputs become ties and therefore affect accuracy and the Cohen kappa score. This matches the reference implementation [here](https://github.com/geoalgo/multilingual-llm-judge/blob/868ff7bba73fc56db7c23cb67d1601531a2dd533/scripts/compute_elo_ratings.py#L24-L45). I kept this behavior to match the reference results, since dropping them or adding an invalid label would change the reported metrics. Let me know if you think we should intentionally diverge.

judgearena/meta_eval/runner.py
ErlisLushtaku: Added it [here](https://github.com/OpenEuroLLM/JudgeArena/commit/03f8e5a54359e91e509b118c64c3c68f4c8877c0).

ErlisLushtaku:
@kargibora Regarding each point:

> `compute_soft_bradley_terry()` reimplements most of `fit_bradley_terry()`.

Good point, I removed it and now use `fit_bradley_terry(..., pref_col="pref_llm")` [here](https://github.com/OpenEuroLLM/JudgeArena/commit/03f8e5a54359e91e509b118c64c3c68f4c8877c0).

> Are these exchangable or we are required to implement this specifically?

`standard` already delegates to `resolve_judge_prompt()`. The remaining modes are not fully exchangeable atm because the shared registry only supports PairScore with `parser_mode="score"`, while Arena-Hard and AlpacaEval need their own verdict parsers. I kept the small meta-eval resolver for these protocol-specific modes.

kargibora:
> standard already delegates to resolve_judge_prompt(). The remaining modes are not fully exchangeable atm because the shared registry only supports PairScore with parser_mode="score", while Arena-Hard and AlpacaEval need their own verdict parsers. I kept the small meta-eval resolver for these protocol-specific modes.

Yes indeed, and to support these frameworks I am experimenting with `Parser` classes for a bit to make this easier.  I will try to take a look at the PR thanks

geoalgo:
I am closing this PR as it is too big and two PRs were open with overlapping focus.

PR #77, opened 2026-07-19
Adopt content-addressed SQLite cache with HF sync

ErlisLushtaku:
## Summary
- Content-address judgements by rendered judge input and isolate configs in hashed SQLite cell folders.
- Wire local store reuse into generate/evaluate and ELO, with fetch-by-default / push-opt-in Hugging Face sync (`judgearena-cache`).
- Add backfill for saved annotation runs and document the shared-cache workflow, including NFS/WAL caveats.


README.md
```diff
@@ -101,6 +101,43 @@ judgearena \
 
 While any key in `--engine_kwargs` is forwarded to the underlying engine (e.g. `vllm.LLM`, `LlamaCpp`, `ChatOpenAI`), existing dedicated flags such as `--max_model_len` and `--chat_template` have higher precedence.
 
+### Shared SQLite cache
+
+Pass `--store_root <path>` to reuse completion and judgement rows across runs.
```
geoalgo: I think adding a high level description before would make sense, something like:

```suggestion
Storing completions and judgement annotations can be made so that LLM calls are avoided by reading from disk. Completions and judgements are written into `~/judgearena-data/db/completions/{task}/{model_name}/{provider}/{config_hash}/` and can be deleted by the user (judgements can be accessed in judgements/). In addition, completions can be fetched from a shared Hugginface (fetching is activated by default, sharing is not).

Pass `--store_root <path>` to reuse completion and judgement rows across runs.
```

scripts/migrate_olmo3_to_store.py
geoalgo, line 1: we may not need that one right?

scripts/benchmark_store_speed.py
```diff
@@ -1,4 +1,4 @@
-"""Benchmark SQLiteCompletionStore / SQLiteJudgementStore load and missing_indices speed.
+"""Benchmark SQLite completion/judgement store lookup speed.
```
geoalgo: Perhaps we move those to `scripts/db/` to avoid spamming `scripts/`?

judgearena/cli.py
```diff
@@ -202,6 +202,12 @@ def _build_elo_args(
         max_model_len=args.max_model_len,
         chat_template=args.chat_template,
         result_folder=args.result_folder,
+        store_root=args.store_root,
```
geoalgo: would it be clear to add those to a namespace with a cache config?

eg to add `args.cache_config` with its own dataclass which contains those 6 fields.

judgearena/estimate_elo_ratings.py
```diff
@@ -157,28 +169,262 @@ def compute_bradley_terry(
     return dict(pd.Series(elo_scores, index=models.index))
 
 
-def _store_folder(store_root: str, kind: str, task: str, model_spec: str) -> Path:
-    provider, model_path = model_spec.split("/", 1)
-    model_name = model_path.replace("/", "--")
-    return Path(store_root) / kind / task / model_name / provider
+def _cache_stores(
+    args: CliEloArgs,
+) -> tuple[
```
geoalgo: We almost never wants to return a tuple with 5 elements. 
Ideally, we splits into subfunctions which do different things, if not possible we should do a dataclasses (but here it really looks like we could do separate functions, for instance, it would be cleaner to pass `judge_system_prompt, judge_user_prompt` as input and we have already 2 outputs less.

judgearena/estimate_elo_ratings.py
```diff
@@ -157,28 +169,262 @@ def compute_bradley_terry(
...
+        judge_user_prompt,
+    )
 
-def main(args: CliEloArgs) -> dict:
-    from judgearena.store_sqlite import SQLiteCompletionStore, SQLiteJudgementStore
 
-    if args.store_root is not None:
-        comp_folder = _store_folder(
-            args.store_root, "completions", args.arena, args.model
+def _collapse_elo_prefs(prefs: list, n_battles: int) -> list:
```
geoalgo:
```suggestion
def _collapse_elo_prefs(prefs[float]: list, n_battles: int) -> list[float]:
```

?

judgearena/estimate_elo_ratings.py
```diff
@@ -157,28 +169,262 @@ def compute_bradley_terry(
...
+    if len(prefs) == n_battles:
```
geoalgo: The contract of this function is a bit weird, it feels to me that this would be cleaner:

```python
# in the caller
if args.swap_mode:
  # average the preferences which are the concatentation of preference with A and B as model and then reverse
  prefs = _collapse_pairwise_prefs(raw_prefs)
else: 
  prefs = raw_prefs 
```

judgearena/estimate_elo_ratings.py
```diff
@@ -157,28 +169,262 @@ def compute_bradley_terry(
...
+        values = [
+            value
+            for value in (forward, reversed_value)
+            if value is not None and not pd.isna(value)
+        ]
+        collapsed.append(sum(values) / len(values) if values else float("nan"))
+    return collapsed
+
+
+def _cache_elo_direction(
```
geoalgo: The name is not super telling, perhaps `_cache_elo_judgement`?

geoalgo (commented):
Thanks for doing this @ErlisLushtaku ! Just had a first read, did not finish it, I will finish the PR in a second pass but already have some comments.

One thing that occurs to me as well, the storage seems to be a bit leaking in the sense that all mains needs to be modified and it seems that we must make them aware of the caching.

However, ideally I think we could have something which is more like:


```
completion_fun = make_model(model, cache_args)
annotation_fun = make_model(judge_model, cache_args)
```

and then call those in any entrypoint with something like

```
completions, metadata = completion_fun(
  inputs=...,
  hash=...,
)
```

which would feel less invasive and hopefully will reduce the LOC change.

geoalgo:
I am closing this PR as it is too big (7K LOC), two PRs were open with overlapping focus and feedback was not adressed so it is not clear if it should be reviewed or not.

We should not open PR with >500 LOC unless clearly motivating it and when doing WIP PR I think it is also important to make clear if it is ready to review or not (as I spend quite some time reviewing 6K LOC which appears to be unused).

PR #78, opened 2026-07-20
Make run metadata compact, explicit, and more reusable

kargibora:
## Summary
Our first draft of metadata is a bit outdated:
- What we save is not checked properly. Quite complicated to put "what to save" 
- We repeat some fields like config, we also save `config.yaml` which makes the run already re-usable. Thus we should remove this duplication
- There were some bugs like metadata capturing what exist in the folder as artifact, however it should track _what it saves_

PR Priority 🟢: (not urgent)

This PR replaces the free-form `run-metadata.v1.json` manifest with a typed and versioned `run-metadata.v2.json` schema shared by the pairwise, ELO, and MT-Bench pipelines.

The saved `config.yaml` remains the source of truth for configuration. The metadata manifest provides a readable run identity, reproducibility fingerprints, execution information, and an explicit inventory of produced artifacts without copying full configuration or result payloads.

## Before and after

| Concern | v1 | v2 |
|---|---|---|
| Configuration | Complete configuration copied into `run` | `config.yaml` referenced by path and SHA-256 |
| Results | Summary copied into metadata | Kept in the result artifact |
| Inputs | Counts generated mechanically from payload keys | Relevant revisions, counts, and stable fingerprints |
| Prompts | Flat optional hash fields | Typed resolved prompt and named MT-Bench variants |
| Artifacts | Output directory scanned recursively | Produced files registered explicitly |
| Artifact integrity | Path and size only | Kind, relative path, size, and SHA-256 |
| Ownership | Free-form dictionaries in each entry point | Shared Pydantic models and writer |

Abbreviated v1 manifest:

```json
{
  "run": {"task": "alpaca-eval", "model": {"name": "..."}},
  "results": {"winrate": 0.5},
  "dataset_statistics": {
    "instructions_count": 100,
    "completions_A_count": 100,
    "completions_B_count": 100
  },
  "artifacts": [
    {"path": "results.json", "size_bytes": 123}
  ]
}
```

Abbreviated v2 manifest:

```json
{
  "schema_version": "judgearena-run-metadata/v2",
  "identity": {
    "workflow": "pairwise",
    "task": "alpaca-eval",
    "model": {"name": "...", "baseline": "..."},
    "judge": {"model": "..."}
  },
  "configuration": {
    "path": "config.yaml",
    "sha256": "..."
  },
  "inputs": {
    "example_count": 100,
    "example_ids_sha256": "...",
    "content_sha256": "..."
  },
  "prompt": {
    "preset": "default",
    "user_sha256": "..."
  },
  "artifacts": [
    {
      "kind": "results",
      "path": "results.json",
      "size_bytes": 123,
      "sha256": "..."
    }
  ]
}
```

## Implementation

- Adds typed metadata models in `judgearena/repro.py`. Unknown fields are rejected and unavailable optional fields are omitted.
- Keeps a minimal identity using the original `RunConfig` field names.
- Records only dataset revisions relevant to the selected task or arena.
- Fingerprints the normalized example-ID set and exact ordered judging inputs without embedding instructions or completions.
- Records judgment and calibration counts only when they add information.
- Fingerprints the exact resolved judge prompt. MT-Bench returns its dynamically selected prompt variants explicitly instead of metadata inferring them from annotation columns.
- Registers produced artifacts explicitly and records their kind, relative path, size, and checksum. Files left by previous runs are not included.
- Captures the Git commit and dirty state, Python and dependency versions, entry point, timestamps, and duration centrally.

All active workflows use the same API:

```python
write_run_metadata(
    output_dir=result_dir,
    entrypoint="judgearena.generate_and_evaluate.main",
    context=RunContext.from_config(
        cfg,
        workflow="pairwise",
        configuration_path=config_path,
    ),
    inputs=InputMetadata.capture(...),
    judge_prompt=resolved_prompt,
    artifacts={...},
    started_at_utc=run_started_at,
)
```

## Adding a field

1. Do not copy information already owned by `config.yaml`, result files, or annotation artifacts.
2. Add the field to its owning typed model, such as `RunIdentity`, `InputMetadata`, or `PromptMetadata`.
3. Prefer an optional field unless every workflow can provide it.
4. Populate it at the workflow boundary and add a focused test in `tests/test_repro.py`. Add entry-point tests only when new plumbing is required.

New output files normally require only another entry in the `artifacts` mapping, not a schema change.

## Compatibility

The filename and schema version intentionally change from v1 to v2. Consumers should branch on `schema_version`. Evaluation results, annotations, configuration files, and active CLI workflows are otherwise unchanged.

## Validation

```bash
uv run pytest
```

- Ruff passed.
- 234 tests passed.

### DIFF
--- a/judgearena/dataset_revisions.py
+++ b/judgearena/dataset_revisions.py
@@ -54,11 +54,51 @@ def hf_revision(repo_id: str) -> str | None:
 def all_dataset_revisions() -> dict[str, str | None]:
     """Return a copy of every pin recorded in this module.
 
-    Used by :func:`judgearena.repro.write_run_metadata` to record the
-    pin table alongside each run so future readers know which version of
-    the data was visible at the time of the run.
+    Run metadata uses :func:`revisions_for_task` or
+    :func:`revisions_for_arena` so it records only sources relevant to that
+    experiment rather than copying the complete global pin table.
     """
     return {
         **HF_DATASET_REVISIONS,
         **{f"raw:{k}": v for k, v in RAW_URL_REVISIONS.items()},
+    }
+
+
+def revisions_for_task(task: str) -> dict[str, str | None]:
+    """Return only the pinned sources used by one instruction benchmark."""
+    repo_ids: list[str] = []
+    if task == "alpaca-eval":
+        repo_ids = ["judge-arena/judge-arena-dataset"]
+    elif task.startswith("arena-hard-"):
+        repo_ids = ["lmarena-ai/arena-hard-auto"]
+    elif task.startswith("m-arena-hard-v0.1"):
+        repo_ids = ["CohereLabs/m-ArenaHard"]
+    elif task.startswith("m-arena-hard-v2.0"):
+        repo_ids = ["CohereLabs/m-ArenaHard-v2.0"]
+    elif task == "mt-bench":
+        repo_ids = ["lmsys/mt-bench"]
+    elif task.startswith("fluency-"):
+        repo_ids = ["geoalgo/multilingual-contexts-to-be-completed"]
+
+    revisions = {repo_id: hf_revision(repo_id) for repo_id in repo_ids}
+    if task == "mt-bench":
+        revisions["raw:lm-sys/FastChat"] = RAW_URL_REVISIONS.get("lm-sys/FastChat")
+    return revisions
+
+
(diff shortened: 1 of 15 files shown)


kargibora:
I am closing this for now as its not a high priority. I feel like we need this but not for now and possible after we define tasks

PR #80, opened 2026-07-21
(2/n) Organize Benchmark Execution Into Responsibility-Owned Packages

kargibora:
## Summary

  The previous PR introduced a shared benchmark runner interface, but benchmark implementations and related utilities are still distributed across the root of `judgearena`.

  For example, pairwise evaluation, ELO execution, MT-Bench, datasets, and run metadata currently live in unrelated root-
  level modules. This makes ownership unclear and increases the number of unrelated files that must be changed when adding or modifying a benchmark.

  This PR reorganizes the existing implementation into responsibility-owned packages and moves benchmark dispatch out of the pairwise runner.

Take a look at [SoP](https://en.wikipedia.org/wiki/Single-responsibility_principle) to understand the reason and shortcomings of the current approach. 

  ## What changed

  Benchmark implementations are now grouped under:

  ```text
  judgearena/benchmarks/
  ├── elo/
  ├── mt_bench/
  ├── pairwise/
  ├── execution.py
  ├── registry.py
  └── runner.py

  Other shared components are also separated:

  judgearena/
  ├── artifacts/
  │   ├── metadata.py
  │   └── run.py
  └── datasets/
      ├── arena_hard.py
      ├── m_arenahard.py
      └── mt_bench.py
```
  Important changes include:

  - Moves pairwise, MT-Bench, and ELO implementations into dedicated packages.
  - Adds a benchmark-owned registry and thin shared dispatcher.
  - Keeps the CLI responsible only for configuration and top-level dispatch.
  - Separates ELO rating calculations from the ELO execution workflow.
  - Centralizes result-directory creation and safe metadata writing.
  - Moves dataset loading utilities under judgearena/datasets.
  - Updates imports and tests to follow the new package ownership.

  Most of the diff consists of file moves and import updates. Existing benchmark protocols, prompts, configuration, and
  evaluation behavior are preserved.

  ## Why this helps

  A benchmark can now own its runner and utilities without adding more logic to the pairwise implementation or the package
  root. It also gives datasets and artifacts stable locations that can be reused by future benchmark integrations.

## Notes
- As can be seen on diffs, most of the LoC coming from moving codes from certain scripts so each script is more readable and does one job.

### CODE REVIEW

File: judgearena/artifacts/run.py [hunk 1]
```diff
@@ -0,0 +1,45 @@
...
+if TYPE_CHECKING:
+    from judgearena.config import RunConfig
+
+logger = get_logger(__name__)
+
+
+def slugify(value: str) -> str:
+    """Return a filesystem-safe model, task, or benchmark name."""
+    slug = re.sub(r"[^a-z0-9]+", "-", value.lower()).strip("-")
+    return slug or "value"
```
geoalgo: not sure about the name (I only vaguely could guess from the name).
Given that we will have it at a bunch of place perhaps we can choose a more telling name like `safe_string`? or `safe_filename`?

judgearena/artifacts/run.py
```diff
@@ -0,0 +1,45 @@
...
+    if attach_log and not cfg.run.no_log_file:
+        attach_file_handler(make_run_log_path(output_dir))
+
+    from judgearena.config import dump_config
+
+    dump_config(cfg, output_dir / "config.yaml")
+    return output_dir
+
+
+def write_run_metadata_safely(**kwargs):
```
geoalgo, line 39: I know it was there before but I am not sure we want to recover from this, if the file-system is not available it seems like this is something that should be fixed and failing is OK.

geoalgo (commented):
LGTM, I have only minor comments.

judgearena/artifacts/run.py
kargibora: However we spend some compute (or money) to generate the results, I think we should save it definitely (or at least give a warning and save it to a temp place).

judgearena/artifacts/run.py
kargibora: I agree.  Slugify is a technical term for generating a safe name from a string with spaces however we dont have to set it up like that. `safe_filename` sounds better to me

kargibora: [61e23f2](https://github.com/OpenEuroLLM/JudgeArena/pull/80/commits/61e23f297e12e1287026408edea294f05d39728f) renames the function to `safe_filename`

judgearena/artifacts/run.py
geoalgo: Either way is fine with me but returning without errors whereas no metadata can be written is potentially bad as well (also the compute cost are usually moderate for evaluations)

geoalgo (approved):
LGTM, we just need to fix ruff / CI before merging, feel free to merge once this is done.

PR #81, opened 2026-07-21
(3/n)  Introduce declarative YAML task definitions

kargibora:
## Summary

  Task-specific configuration is currently distributed across Python constants, dataset loaders, baseline mappings, prompt
  registries, and runner conditions. Adding or changing a task therefore requires modifying several unrelated modules.

  This PR introduces declarative YAML task definitions, inspired by the task packaging approach used by [lm-evaluation-
  harness](https://github.com/EleutherAI/lm-evaluation-harness/tree/main/lm_eval/tasks/gsm_plus).

  AlpacaEval is migrated as the first packaged task. Existing benchmarks continue using their current fallback behavior
  until they are migrated in later PRs.

  ## Task definitions

  A task YAML describes the stable contract of a benchmark:

  ```yaml
  task: alpaca-eval
  task_version: 1

  dataset:
    adapter: judgearena_tables
    sources: ...
    fields: ...

  protocol:
    runner: pairwise
    generation: ...
    baseline: ...
    judge: ...
    scoring: ...
  ```
  The main sections are:

  - **task**: stable task identity, version, description, and tags.
  - **dataset**: source locations, pinned revisions, loader adapter, and canonical field mappings.
  - **protocol**: how the task is generated, judged, and scored.
  - **metadata**: links to the reference implementation or paper.

  The protocol does not contain implementation code. It references registered components such as **pairwise**,
  **judgearena_tables**, **pairwise_preference**, and **pairwise_win_rate**.

  ## Main components

  The new task package contains three main responsibilities:
  ```
  judgearena/tasks/
  ├── definitions/   # YAML task definitions
  ├── schema.py      # Valid task structure
  ├── loader.py      # YAML loading and inheritance
  ├── registry.py    # Task discovery and component validation
  └── cli.py         # list, show, and validate commands
  ```
  - The **schema** defines the supported dataset sources, baseline strategies, judge settings, and scoring configuration.
    Unknown fields are rejected.

  - The **loader** reads YAML safely, resolves optional _base.yaml inheritance, and calculates hashes for the source and
    resolved definitions.

  - The **registry** discovers packaged tasks by ID and verifies that every referenced runner, dataset adapter, prompt,
    parser, and scorer exists.

  - The **CLI** allows users to inspect and validate task definitions before running them.

  ## Runtime behavior

  When a task is executed, JudgeArena resolves it through the registry:

  Task ID
      -> load and validate task YAML
      -> resolve dataset source and field mappings
      -> apply task protocol defaults
      -> select the registered runner
      -> generate, judge, and score

  Task definitions provide benchmark defaults, while users can still configure experiment-specific values.

  The effective precedence is:

  CLI flags -> run config YAML -> task-defined defaults -> framework defaults

  For example:
  - The task selects the runner and dataset adapter.
  - The task provides the default baseline, prompt, swap mode, and optional judge temperature.
  - --model.baseline may override the task baseline when the task permits it.
  - Explicit judge settings from the CLI or run config remain unchanged.
  - Stable task identity remains in the task YAML, while model and judge choices remain run-level configuration.

  ## Additional changes

  - AlpacaEval dataset loading now uses the source revision and field mappings declared in its task YAML.
  - Task definitions are included in the installed Python package.
  - Run metadata records the task version and resolved definition hash without copying the complete YAML into every
    result.
  - Legacy fallback behavior is preserved for tasks that have not yet been packaged.


judgearena/cli.py
```diff
@@ -27,8 +28,15 @@ def _format_config_error(exc: ValidationError) -> str:
...
+    args = list(sys.argv[1:] if argv is None else argv)
+    if args[:1] == ["tasks"]:
+        from judgearena.tasks.cli import run_task_command
+
+        run_task_command(args[1:])
+        return
+
     try:
-        cfg = build_run_config(argv)
+        cfg = build_run_config(args)
```
geoalgo: Can you add a comment here on the rationale? Why doing run_task_command only when "tasks" is passed as CLI?
It is not obvious to understand what this code does.

judgearena/paths.py
```diff
@@ -32,7 +32,16 @@ def _data_root_path() -> Path:
...
+    """Download a registered or legacy instruction-table dataset."""
+    from judgearena.tasks.registry import get_packaged_task
+
+    resolved_task = get_packaged_task(name)
+    if resolved_task is not None:
+        from judgearena.datasets.judgearena_tables import download_task_sources
+
+        download_task_sources(resolved_task, local_path)
+        return
+
```
geoalgo: It is not cleaer what packaging tasks are given this code and what are the changes.
Can we have a single downloading function?

If keep it, it feels to me that else would be both cleaner and easier to read here.

```suggestion
    if resolved_task is not None:
        from judgearena.datasets.judgearena_tables import download_task_sources

        download_task_sources(resolved_task, local_path)
    else:
```

judgearena/tasks/registry.py
```diff
@@ -0,0 +1,140 @@
...
+        *,
+        adapters: AdapterCatalog | None = None,
+    ):
+        if definitions_root is None:
+            definitions_root = files("judgearena.tasks").joinpath("definitions")
+        self._loader = TaskLoader(definitions_root)
+        self._adapters = adapters or AdapterCatalog()
+        self._tasks: dict[str, ResolvedTaskSpec] | None = None
+
+    def get(self, task_id: str) -> ResolvedTaskSpec:
```
geoalgo: I think it would be better to avoid a registry that basically reimplement most dictionary functions.
We can get that by having:
`load_tasks(...) -> dict[str, ResolvedTaskSpec]`
and just dealing with the dictionary in downstream code.

We save 100 LOC and also the need to check an API which is really just a dict interface.

judgearena/paths.py
```diff
@@ -32,7 +32,16 @@ def _data_root_path() -> Path:
 
 
 def download_hf(name: str, local_path: Path) -> None:
```
geoalgo: We have multiple download_hf it seems... 
Can you add a TODO to merge to one if not done in this PR?

judgearena/tasks/loader.py
```diff
@@ -0,0 +1,231 @@
...
+        if value is None:
+            merged.pop(key, None)
+        elif isinstance(value, dict) and isinstance(merged.get(key), dict):
+            merged[key] = _merge_mapping(merged[key], value)
+        else:
+            merged[key] = deepcopy(value)
+    return merged
+
+
+class TaskLoader:
```
geoalgo, line 106: See my other comment, we just need one function to load all tasks which returns a dict all of those can be in the same file.

judgearena/tasks/registry.py
```diff
@@ -0,0 +1,140 @@
...
+
+@dataclass(frozen=True)
+class ValidationReport:
+    """Summary returned after validating every packaged task."""
+
+    tasks: tuple[TaskSummary, ...]
+
+    @property
+    def count(self) -> int:
+        return len(self.tasks)
```
geoalgo: This one feels unnecessary to me, we are not calling count, we can just provide the type without providing this extra type.

judgearena/tasks/cli.py
```diff
@@ -0,0 +1,58 @@
+"""Implement ``judgearena tasks list|show|validate`` inspection commands."""
```
geoalgo, line 1: Perhaps task_cli.py would be clearer?

judgearena/tasks/definitions/alpaca_eval/alpaca-eval.yaml
```diff
@@ -0,0 +1,38 @@
+schema_version: 1
```
geoalgo, line 1: Where is the schema of tasks leaving? 
This one we would really one a dataclass or custom type in python code.

judgearena/datasets/judgearena_tables.py
```diff
@@ -0,0 +1,70 @@
...
+
+from pathlib import Path
+
+import pandas as pd
+from huggingface_hub import snapshot_download
+
+from judgearena.tasks.schema import HuggingFaceDatasetSource, ResolvedTaskSpec
+
+
+def download_task_sources(task: ResolvedTaskSpec, local_dir: Path) -> None:
```
geoalgo, line 13: should we keep only one from download_all and this one?

geoalgo (commented):
LGTM overall, I have left some comments.

judgearena/cli.py
kargibora: Hi David,
This function basically changes the execution flow for the main entry point. If `judgearena tasks` if provided we can execute arbitrary command like `--list` to see all the available tasks (or validate any) - if not provided simply execute the workflow. I think we can discard this if you think we dont need such functionality but for completeness I did implement it anyways.

judgearena/paths.py
kargibora: `download_df` in `paths.py` is indeed a dead code. I will remove this stale code (we are importing from utils always).

judgearena/paths.py
kargibora: Currently our `download_hf` is in its respective `io.py` file, however the code is exact so I will comment it here.

Packaged tasks are basically the tasks that we have implemented with this new `tasks.yaml` pipeline. As I did not change everything and all the datasets are added step-by-step as a task, this allows to keep downloading datasets in a legacy way before they are added (if we merge all stacked PR's, this will removed for example. I tried to make it so each seperate PR still works for all the tasks even if they did not converted to a YAML file yet).

I agree on `else:` branch, it makes it easier to read.

judgearena/tasks/registry.py
kargibora: I agree, this is an over-complication that is added and can easily be avoided. If we ever need such functionality we can change it without a problem in the future. I will change it

judgearena/tasks/cli.py
kargibora: I think we can put this in two possible paths:
- Under `judgearena.task_cli.py`
- Current

I would prefer the current version, as this makes all the tasks-related codes bundled in a package.  `tasks.task_cli.py` is also feels a bit redundant in terms of wording. However happy to rename it if you still find it confusing to have two seperate `cli.py`

judgearena/tasks/definitions/alpaca_eval/alpaca-eval.yaml
kargibora: `judgearena/tasks/schema.py` rejects the unknown keys and it uses a special class for validating the yaml tasks. Does this clarifies this? 

We also define it with
`TaskSpec.schema_version: Literal[1]`, so if we decide on changing some important stuff in the pipeline, `task validate` fails for the tasks with older versions.  If this is over-complication can also make it better

judgearena/datasets/judgearena_tables.py
kargibora: `download_task_sources` is an adapter from newer version of reading revision from `tasks.yaml` and dowloading the specified task. IT is basically used in `download_hf` which then `download_all` calls as:
```py
        if is_arena_hard_dataset(dataset):
            download_arena_hard(dataset=dataset, local_tables_path=local_path_tables)
        else:
            download_hf(name=dataset, local_path=local_path_tables)
```
So I think at this point we should not consodliate it in the current PR. As we add more tasks and complete the entire stacked PRs where we also done with adding the arena as tasks (up to #86 ), than a single function can download the entire available tasks, as well as specific ones (also another advantage of our classed approach, now we wont be needing if statements as a class will handle the downloading by itself). 

My suggestion is to let it stay like this at the current PR, and let the other PR's in the stack handles this change after we added all the tasks.

judgearena/tasks/registry.py
kargibora: Agreed, removed in the commits

judgearena/tasks/loader.py
kargibora: I thing merging these two into a single `registry` would make the `registry` larger, as function wise, `registry.py` handles resolving the tasks by calling the `loader`. However as `registry` is the only script that calls `TaskLoader`, it is also OK to put them all in. 

However, compared to `TaskRegistry` which is basically a dict, I feel like `TaskLoader` class is more graceful. Instead of providing the functions like `load,..., discover`, it just provides a single object `loader` which just handles all the things we want to do.

Happy to convert it to functions or merge it with `registry` if you think that is more cleaner and readable approach

judgearena/cli.py
geoalgo: Ok I would suggest something like this to improve readability:

```python
is_task_cli = args[:1] == ["tasks"]  
if is_task_cli:
    # handle the case where we run `judgearena tasks ARGS`
    from judgearena.tasks.cli import run_task_command
    run_task_command(args[1:])
else:
    # generic case for other commands
    cfg = build_run_config(args)
```

which is more readable (but still a bit hacky).

judgearena/tasks/definitions/alpaca_eval/alpaca-eval.yaml
geoalgo: (sorry I meant "living" but it seems you got what I meant)

> judgearena/tasks/schema.py

ok makes sense, I was wondering if it is explicit somewhere and it seems so (although perhaps currently not super well flatten-out for a new user to understand, but we can have a YAML example which details all field with a comment or something like this in the doc)

judgearena/datasets/judgearena_tables.py
geoalgo: Makes sense, perhaps we just add support later to define a custom download function in case its needed when we cant rely on just dowload_hf.

judgearena/tasks/loader.py
geoalgo: Why do we need both? Both seems to be utils to load tasks?
It seems that this class has complex logic to detect whether tasks are valid after a recursive walk.

I think we can make it much simpler, just browse for files with exact names "task_config.yaml" in subdirs.
We do not want to do any patching, anything that does not match convention should just be rejected with a warning possibly.

judgearena/tasks/definitions/alpaca_eval/alpaca-eval.yaml
kargibora: Yes totally agree, later PR also adds the markdown doc under `judgearena/tasks` hopefully that will resolve this

judgearena/tasks/loader.py
kargibora: So the catch is future PR's introduces some smarter logic for loading and extending tasks. For example specifying `xxx-en` will be load the english dataset or some datasets like `official protocol configs` we are willing to add extendes the `base.yaml.

judgearena/cli.py
kargibora: I agree, this is more readable

judgearena/tasks/loader.py
geoalgo: sorry cant parse your sentence.

judgearena/cli.py
```diff
@@ -27,8 +28,21 @@ def _format_config_error(exc: ValidationError) -> str:
...
+    # instead of running an evaluation, so it has its own argparse grammar (see
+    # tasks/cli.py) and must be routed before build_run_config, which only parses
+    # run/eval flags and would reject the subcommand form.
+    is_task_cli = args[:1] == ["tasks"]
+    if is_task_cli:
+        from judgearena.tasks.cli import run_task_command
+
+        configure_logging()
+        run_task_command(args[1:])
+        return
```
geoalgo: we still have return, can you do `else` as described in the comment bellow?

judgearena/cli.py
```diff
@@ -27,8 +28,21 @@ def _format_config_error(exc: ValidationError) -> str:
...
+    args = list(sys.argv[1:] if argv is None else argv)
+    # `judgearena tasks {list,show,validate}` inspects packaged task definitions
+    # instead of running an evaluation, so it has its own argparse grammar (see
+    # tasks/cli.py) and must be routed before build_run_config, which only parses
+    # run/eval flags and would reject the subcommand form.
+    is_task_cli = args[:1] == ["tasks"]
+    if is_task_cli:
+        from judgearena.tasks.cli import run_task_command
+
+        configure_logging()
```
geoalgo: why would we configure logging only in this case? dont we want to have it before the if?

geoalgo (approved):
approving to not block the other PRs, doing the else would be better, I could not parse one of your sentence.

judgearena/cli.py
kargibora: So there was a bit confusion.. Because `run_task_command` basically executes `task/cli.py`, which has logging but in `cli.py`, we configure logging using `verbosity` and `log_file` for the main executions. However we actually do not need logging for the task commands so I think its safer to simply remove it

judgearena/cli.py
kargibora: Fixed it, as there was few blocks I did not put `else:` but it makes it more readable.

judgearena/cli.py
kargibora: Fixed this, we dont need `logging` for task commands which just validates or prints a task. `logging` is useful if we want to have some organized hierarchy in logging as well as a way to save them for debugging & reporting.

judgearena/tasks/loader.py
kargibora: Ah sorry. I got your question wrong. We can indeed just merge these two functionailties, because `TaskLoader` is never called in other places. There should be some complex logic for resolving the task, but it can be handled by the functions instead of some class object. I will simplify it

PR #86, opened 2026-07-21
(7/n) Complete declarative task runtime and shared pairwise orchestration

kargibora:
## Summary

This PR connects packaged task YAML definitions to dataset loading, baseline selection, benchmark execution, and scoring, and splits the task schema into focused modules.

Previously, task definitions described benchmark settings, but parts of the runtime still selected behavior through hard-coded branches, repeated registry lookups, or direct function calls. Adding another task could therefore require changes across several unrelated files.

After this change, a task is resolved once and passed through the pairwise evaluation pipeline. Existing components can be reused by referencing their registered names from task YAML.

## Before and after

```
 Before                                                       After
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 Task information was partly defined in YAML and partly       Stable task behavior is defined in task YAML and
 selected in Python.                                          validated by typed schemas.
───────────────────────────────────────────────────────────  ───────────────────────────────────────────────────────────
 Runners and dataset loaders could resolve the same task      The dispatcher resolves the task once and passes the
 repeatedly.                                                  resulting specification downstream.
───────────────────────────────────────────────────────────  ───────────────────────────────────────────────────────────
 Pairwise tasks relied on benchmark-specific routing and      Compatible tasks share the same pairwise runner and
 duplicated setup logic.                                      dataset orchestration.
───────────────────────────────────────────────────────────  ───────────────────────────────────────────────────────────
 Scoring configuration was validated but not always used      The runner resolves and executes the scoring adapter
 by the runtime.                                              declared by the task.
───────────────────────────────────────────────────────────  ───────────────────────────────────────────────────────────
 Adding fields to one large schema made task definitions      Schemas are separated by responsibility and protocol-
 harder to maintain.                                          specific fields remain isolated.
```

## Architecture

The runtime now follows this path:

```
Task YAML
    ↓
Schema validation and inheritance
    ↓
Resolved task specification
    ↓
Benchmark and dataset registries
    ↓
Shared or protocol-specific runner
    ↓
Generation, judging, scoring, and artifacts
```

The new schema package separates datasets, sources, baselines, pairwise protocols, MT-Bench protocols, and resolved task models. This prevents a single schema file from growing whenever a new benchmark family introduces specialized fields.

Most new tasks do not require a new schema or runner. A task using existing components only needs a YAML definition selecting its dataset adapter, runner, baseline policy, judge protocol, and scorer.

A new Python component is only necessary when a benchmark introduces genuinely different behavior:

- A dataset adapter handles benchmark-specific downloading and normalization.
- A benchmark runner handles a different evaluation workflow.
- A protocol schema validates new task-owned algorithm settings.
- A scoring adapter handles a different result interpretation.

Executable behavior remains in Python rather than YAML; YAML only selects registered and validated components.

## Changes

- Split the task schema into focused modules.
- Added shared pairwise dataset and runner orchestration.
- Passed resolved task specifications through the runtime instead of repeating lookups.
- Made task-defined scoring adapters operational: the runner now resolves the scorer named by the task, and `primary_metric` / `higher_is_better` / `parser` were removed from task YAML since the scorer and prompt preset already own them.
- Added documentation for defining tasks and extending the available components.
- Preserved runtime overrides while keeping stable benchmark identity in task YAML.

## Notes
Most of the changes are coming from splitting `schema.py` to multiple files, thus this is just a refactoring step as next PR's will introduce two more task which can inflate the same file if they require some special fields.


judgearena/tasks/schema/__init__.py
```diff
@@ -0,0 +1,79 @@
+"""Public schema API for declarative task definitions.
```
kargibora, line 1: **Note:**
All these related changes simply splits functions. As `mt_bench` can require some specific fields that others ignore, we only need a shared components + special components on their own file.

judgearena/tasks/schema/baselines.py
```diff
@@ -0,0 +1,43 @@
+"""Baseline-selection policies shared by pairwise protocols."""
```
kargibora, line 1: This was a design problem we have talked before. Baseline is used to determine which model should be used as default - and which tasks requires a baseline provided by the config file we send. We probalby dont need many fields other than the ones that requires us to introduce functionality like using multiple different baselines sucnh as`CategoryDefaultBaseline`. However if you think this is over-engineering happy to remove it

judgearena/benchmarks/mt_bench/runner.py
```diff
@@ -345,7 +348,9 @@ def _run_mt_bench_preset(
     )
 
 
-def run_mt_bench_benchmark(cfg: RunConfig):
+def run_mt_bench_benchmark(
```
geoalgo: This feels slightly weird to me. I would imagine that tasks resolve a run function which gets executed.
Here it feels weird to have `run_mt_bench_benchmark` takes as input `cfg: RunConfig` and `_resolved_task: ResolvedTaskSpec`.

judgearena/benchmarks/pairwise/baselines.py
```diff
@@ -3,9 +3,101 @@
...
+    @classmethod
+    def per_row(cls, series: pd.Series) -> BaselinePlan:
+        return cls(baseline_by_index=series.astype(str).rename("model_B"))
+
+    @property
+    def unique_models(self) -> list[str]:
+        return sorted(self.baseline_by_index.dropna().unique().tolist())
+
+    @property
+    def is_flat(self) -> bool:
```
geoalgo: why `is_flat` rather than `is_unique_model`?

judgearena/tasks/schema/baselines.py
geoalgo: It is not clear why we need it.
Why cant we just have a `baseline: str = "gpt-4o"` config key with a default value in the task configurations?

judgearena/benchmarks/pairwise/scoring.py
```diff
@@ -0,0 +1,46 @@
+"""Runtime scoring adapters for pairwise preference tasks."""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from dataclasses import dataclass
+
+import pandas as pd
+
+from judgearena.utils.eval import PrefSummary, compute_pref_summary
```
geoalgo, line 10: are we using compute_pref_summary anywhere else? if not lets put the code in the object bellow to simplify.

judgearena/benchmarks/pairwise/scoring.py
```diff
@@ -0,0 +1,46 @@
...
+
+    name: str
+    primary_metric: str
+    higher_is_better: bool
+    summarize: Callable[[pd.Series], PrefSummary]
+
+
+_PAIRWISE_SCORERS = {
+    "pairwise_win_rate": PairwiseScorer(
+        name="pairwise_win_rate",
```
geoalgo: DRY, name appears twice. 
Either we keep the dictionary and rename name field (best approach I think) or we have a list of PairwiseScorer.

judgearena/benchmarks/pairwise/scoring.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+}
+
+PAIRWISE_SCORER_NAMES = frozenset(_PAIRWISE_SCORERS)
+
+# Used by legacy tasks that have no YAML definition to declare a scorer.
+DEFAULT_PAIRWISE_SCORER = "pairwise_win_rate"
+
+
+def resolve_pairwise_scorer(name: str) -> PairwiseScorer:
```
geoalgo: why do we need this? it seems that the only benefit is to have a custom ValueError instead of the standard ValueError for a dict.
If so lets remove it and just call `PAIRWISE_SCORERS[name]` in downstream code.

judgearena/datasets/pairwise.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@dataclass(frozen=True)
+class PairwiseTaskData:
+    """Normalized instructions and optional pre-generated model outputs."""
+
+    instructions: pd.DataFrame
+    model_outputs: pd.DataFrame | None = None
+
+    def completions_for(self, model: str) -> pd.Series | None:
```
geoalgo: This naming seems clearer to me:
```suggestion
    def model_completion(self, model: str) -> pd.Series | None:
```

judgearena/datasets/pairwise.py
```diff
@@ -0,0 +1,90 @@
...
+
+logger = get_logger(__name__)
+
+
+@dataclass(frozen=True)
+class PairwiseTaskData:
+    """Normalized instructions and optional pre-generated model outputs."""
+
+    instructions: pd.DataFrame
+    model_outputs: pd.DataFrame | None = None
```
geoalgo, line 27: when is it None? it seems like something unexpected, perhaps we dont support this case which simplifies code

judgearena/datasets/pairwise.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def completions_for(self, model: str) -> pd.Series | None:
+        """Return model completions aligned to the instruction index."""
+        if self.model_outputs is None:
+            return None
+
+        required = {"instruction_index", "model", "output"}
+        missing = sorted(required - set(self.model_outputs.columns))
+        if missing:
+            raise ValueError(
```
geoalgo, line 35: This should not happen when querying the object but when creating it.

judgearena/benchmarks/pairwise/runner.py
```diff
@@ -247,7 +172,14 @@ def _align_completion_series(df: pd.DataFrame) -> pd.Series:
         return df.set_index("instruction_index").loc[instructions.index, "completion"]
 
     def _load_or_generate_completions(model_spec: str, *, role: str) -> pd.Series:
-        preloaded = try_load_dataset_completions(cfg.task, model_spec, n_instructions)
+        if task_data is not None:
+            preloaded = task_data.completions_for(model_spec)
+            if preloaded is not None:
+                return preloaded.loc[instructions.index]
+        else:
+            preloaded = _try_load_legacy_dataset_completions(
```
geoalgo: cant we put the logic above in `_try_load_legacy_dataset_completions`?
(I mean this
        if task_data is not None:
            preloaded = task_data.completions_for(model_spec)
            if preloaded is not None:
                return preloaded.loc[instructions.index]
        else:
)
otherwise the code is weird since the block is also trying to load dataset completions.

judgearena/tasks/schema/sources.py
```diff
@@ -0,0 +1,43 @@
+"""Pinned external resource schemas used by task datasets."""
+
+from typing import Annotated, Literal
+
+from pydantic import Field
+
+from judgearena.tasks.schema.base import StrictFrozenModel
+
+
+class HuggingFaceDatasetSource(StrictFrozenModel):
```
geoalgo, line 10: Do we really need those sources?
Dataset are always loaded locally, we have code to download them so things are reproducible given a git hash.
Can we remove this as a simplification given that the PR is quite complex?

judgearena/tasks/README.md
```diff
@@ -0,0 +1,113 @@
+# Task definitions
```
geoalgo, line 1: I wish I would have started the PR by this 😂.

judgearena/tasks/README.md
```diff
@@ -0,0 +1,113 @@
...
+  baseline policy, judge protocol, and scoring definition.
+- Keep experiment choices in `RunConfig`: candidate and judge models, sampling
+  parameters, instruction limits, caching, and supported overrides.
+- Prefer a new task version when changing data or evaluation behavior. Changing
+  only the model or runtime settings does not create a new task.
+
+## Layout
+
+```text
+tasks/
```
geoalgo, line 16: We do have one folder per task right? (also we should support custom files from the folder so for instance mt_bench.py should probably be in mt_bench/)

judgearena/tasks/schema/baselines.py
kargibora: For `arena-hard-v2.0`, the default officilal implementation uses different prompts and baseline (for creativity at least). If we converge it to a single check like `baseline:...` then we can not achieve this from a task (or any task that requires different baselines).

I think we can just pop out the keys like `runtime_required` and `none` etc. However I think we need a bit more structure than using a simple string:

```yaml
baseline: gpt-4-0314          # string → fixed reference model (5 of 6 tasks)

baseline:                     # mapping with references → per-category
  category_field: category
  references:
    hard_prompt: o3-mini-2025-01-31 # can define different baseliens per category as needed for arena hard v2
```

judgearena/benchmarks/mt_bench/runner.py
kargibora: You are exactly right about this thanks for catching, as `mt_bench` is a specific pipeline for `mt_bench` it was implemneted in a way that it does not require to pass the task as it can just load it (because its unique). So it was a bad implementation.

However I think we need to pass `ResolvedTask` nevertheless, it does not just resolve which functions to execute but also how to execute them (original `mt_bench` has different temperature for different categories for example). The `mt-bench.yaml` provides such informations and it should be resolved and used properly in the respective runner functions

judgearena/benchmarks/pairwise/baselines.py
kargibora: I think `is_single_model` is a bettername. (Also used by the following function). Because `BaselinePlan` calculates which baseline to assign to every row - which can be different as `arena-hard` uses multiple baselines across categories for example.

judgearena/benchmarks/pairwise/scoring.py
kargibora: Yes, three other call sites: 
* _compute_grouped_stats in utils/eval.py (MT-Bench's per-category/per-turn stats)
* The ELO runner,
* evaluate.py. 

This behaviour already existed before in the pipeline and I did not want to simplify it further for the current PRs as we are just adding some functionalities.

judgearena/benchmarks/pairwise/scoring.py
kargibora: Agreed, will pop the `name` as key.

```py
@dataclass(frozen=True)
class PairwiseScorer:

    primary_metric: str
    higher_is_better: bool
    summarize: Callable[[pd.Series], PrefSummary]


_PAIRWISE_SCORERS = {
    "pairwise_win_rate": PairwiseScorer(
        primary_metric="winrate",
        higher_is_better=True,
        summarize=compute_pref_summary,
    )
}
```

judgearena/benchmarks/pairwise/scoring.py
kargibora: Removed, the scorers are already validated with `task.yaml` so we dont need this. Thanks for catching it

judgearena/datasets/pairwise.py
kargibora: This is actually intended.  If `model_outputs` is none, we generate completions in the downstream. For example if we specifiy two models for `alpaca-eval` but `baseline` is not in the table (we may want to compare two checkpoints of our model), we generate completions for both. So this already exist. Maybe I can put more stronger comment:
```py
class PairwiseTaskData:
    """Normalized instructions and optional pre-generated model outputs.

    ``model_outputs`` is ``None`` when the dataset ships no pre-generated
    completions; the runner generates whatever it cannot find here.
    """
```

judgearena/datasets/pairwise.py
kargibora: Agreed, changed it.

judgearena/datasets/pairwise.py
kargibora: Moved it to `__post_init__` so it is checked after the creation. This will allow invalid objects to not be constructed

judgearena/tasks/schema/sources.py
kargibora: The reason sources moved into the task file isn't reproducibility:

(a) the pin becomes required and validated: the old dataset_revisions.py returned None for unlisted repos and silently downloaded latest main. Maybe we can also assign this to some tasks but this would create some divergence that we may not notice

(b) pins become per-task instead of one global revision per repo, which the official protocol need (frozen official data vs newer community data from the same repo). 

(c) a task's data dependencies are readable from its own YAML.

I have removed the unused fields and `LocalSource`; they were here for completeness but we dont need them and hence, YAGNI.

judgearena/tasks/README.md
kargibora: Its one folder per task family (arena hard has two versions, possibly 2 more for pinning official modes that replicates the protocol)

For custom files, agreed but maybe we can do this in a future PR as this PR already got quite complicated (possibly the most complicated one)

judgearena/tasks/schema/baselines.py
geoalgo: I see perhaps we can have a tasks as a fix baseline / judge and support a collection of tasks as we do in oellm-eval. We can leave it like that for now.

judgearena/datasets/pairwise.py
geoalgo: LGTM

geoalgo (approved):

PR #88, opened 2026-07-23
Docs/lumi setup

kargibora:
# Summary
This PR creates and updates some documentary for creating a container for JudgeArena as well as how to use it in cluster-environment (with some specific details for oellm-eval). 

It also changes:
- HF_HOME can be used to cache datasets - JUDGEARENA_DATA is not required. 
- More default configs are constructed, also shows an how to construct a proper config to the users
- EloReport is updated for showcasing which arena dataset we have used. 
- Downloading is wired to CLI so it is easy to download a newer dataset with commands like `judgearena-download alpaca-eval elo-lmarena-100k` making it much easier to setup the first steps


geoalgo (commented):
This is coupled with changes on task, can we have a separate for task definition?

Also I think we may actually prefer to have this in oellm-eval since we have container descriptions there so it feels a better place?

kargibora:
We can put oellm-eval things to oellm-eval, but container information can be useful (or should we assume ppl can already do this?)

kargibora:
Closing this

PR #92, opened 2026-08-04
(8/n) Add ELO/Battles datasets as Task

kargibora:
## Summary

This PR packages the ELO/arena benchmarks as declarative tasks, unifying them with the rest of the task system. ELO runs are now selected, validated, and configured the same way as every other benchmark instead of through special-cased code paths.

Previously, ELO was dispatched by matching a task-name prefix, arena datasets were downloaded through their own hardcoded path, and arena choices and dataset revisions lived in Python constants. Adding or changing an arena meant touching several unrelated files.

After this change, each arena is a task definition: `elo-lmarena`, `elo-lmarena-100k`, `elo-lmarena-140k`, and `elo-comparia`. The task YAML declares the arena, its pinned data sources, and its scoring defaults; the runtime resolves the task once and routes it through the shared benchmark registry.

## What changed

- **ELO protocol schema.** Task YAML can now declare an `elo` protocol (arena, judging, Bradley-Terry scoring defaults), validated like the pairwise and MT-Bench protocols.
- **Arena battles as a dataset kind.** Battle datasets (model A vs. model B with a human winner) are a different contract than instruction datasets, so they get their own small adapter type. ELO tasks download and load their arena data through the same dataset registry as everything else.
- **One dispatch path.** The task-name-prefix check for ELO is gone; the ELO runner is a regular registered benchmark.
- **Pins moved into task YAML.** Arena dataset revisions previously kept in `dataset_revisions.py` now live with the task that uses them.
- **Config validation.** Unknown task names are rejected at config time with a pointer to `judgearena tasks list`; ELO runtime settings (battle sampling, calibration, bootstraps) stay in the run config, while stable arena identity stays in the task YAML.


judgearena/benchmarks/elo/scoring.py
```diff
@@ -0,0 +1,28 @@
+"""Runtime scoring adapters for ELO rating tasks.
```
kargibora, line 1: Note: Although it seems like this only provides a single function, scoreres are generally needed for implementing specific behaviours defined in official protocls like length-controlled results.

judgearena/benchmarks/elo/runner.py
```diff
@@ -438,7 +447,6 @@ def run_judge() -> pd.DataFrame:
 
             judge_chat_model_cal = make_model(
                 model=cfg.judge.model,
-                max_tokens=cfg.judge.max_out_tokens,
```
kargibora, line 441: @geoalgo  this was a bug introduced by some PR in main. I actually don't know when this is introduced (possibly alex's PR). Unfortunutely it was breaking the execution.

judgearena/config.py
```diff
@@ -380,58 +381,76 @@ class RunConfig(BaseSettings):
...
 
     elo: EloArgs | None = None
-    """ELO-task settings (only for ``elo-*`` tasks)."""
+    """Runtime settings used only by tasks with an ELO protocol."""
 
     run: RunArgs = Field(default_factory=RunArgs)
     """Run-level settings (seed, output, caching, logging)."""
 
     @model_validator(mode="after")
     def _validate(self) -> RunConfig:
```
kargibora, line 390: This is getting more complicated, which basically validates the resolved config file we pass for succesful execution. Might have to find better solution in future (not related to this PR)

judgearena/benchmarks/elo/runner.py
```diff
@@ -135,14 +137,21 @@ def render(self) -> None:
...
+    arena = protocol.arena
+    scorer = ELO_SCORERS[protocol.scoring.adapter]
     run_started_at = datetime.now(UTC)
     rng = np.random.default_rng(cfg.run.seed)
 
     # Step 1: Load arena battles
-    logger.info("Step 1: Loading battles from %s", cfg.elo.arena)
-    df_arena_all = load_arena_dataframe(arena=cfg.elo.arena)
+    logger.info("Step 1: Loading battles from %s", arena)
+    df_arena_all = load_battles(task)
```
geoalgo, line 154: Not necessarily for now but dont we want to have each task define a `load_data` `run_annotations` `post_process` `report` which are called by a common main?

judgearena/benchmarks/elo/runner.py
geoalgo: Thanks for catching this could you add a test for this?

judgearena/tasks/definitions/elo/elo-lmarena-100k.yaml
```diff
@@ -0,0 +1,17 @@
+extends: _base.yaml
+task: elo-lmarena-100k
+description: Arena-anchored ELO evaluation using LMSYS Chatbot Arena 100k battles.
+
+dataset:
```
geoalgo, line 9: how do we support subsetting different languages here?

judgearena/config.py
```diff
@@ -380,58 +381,76 @@ class RunConfig(BaseSettings):
...
-                raise ValueError(
-                    f"model.baseline cannot override the baseline defined by "
-                    f"task {self.task!r}."
+                    f"Unknown task {self.task!r}; use 'judgearena tasks list' to "
+                    "inspect packaged tasks."
                 )
+            # Fluency tasks are not packaged yet and run through the legacy path.
+            if self.elo is not None:
+                raise ValueError("elo config is only valid for ELO tasks.")
+            if self.model.name is None:
```
geoalgo, line 401: cant we add a general `validate_config` in tasks which always called in a common main? (not for this PR but as potential improvement)

geoalgo (approved):

judgearena/benchmarks/elo/runner.py
kargibora: Yes of course, I will add it

judgearena/tasks/definitions/elo/elo-lmarena-100k.yaml
kargibora: Same as we did in others, we can put a `resolver` that takes a look at column of the dataset `lang` and filters the samples. Should be a few line change (I will add it)

judgearena/benchmarks/elo/runner.py
kargibora: This is a nice question, as PR's are already large (and we have enough tool to just build this) this can be doable. We can discuss this perhaps? (Currently the pipelines are almost same, just dispatched in their own runners with more task configuration. Next steps would be to decompose pipelines into more repeatable steps)

judgearena/tasks/definitions/elo/elo-lmarena-100k.yaml
kargibora: I have added the support. Now we can run `xxx-en` to select languages as well for each of the arena

PR #93, opened 2026-08-04
(9/9) Fluency Task Declarations

kargibora:
## Summary

This PR packages the fluency benchmark as a declarative task; the last benchmark to migrate. With every dataset now owned by a task definition, it also deletes all the legacy code paths that existed only for unpackaged tasks. This completes the task-YAML stack.

## Fluency as a packaged task

- One `fluency` task family with 43 language variants (`fluency-french`, `fluency-mandarin-chinese`, …), resolved through the same suffix-variant mechanism as m-ArenaHard.
- The `multilingual-fluency` dataset source and revision are pinned in the task YAML and loaded through a registered `fluency` dataset adapter.
- New `base_completion` generation mode: the task protocol now declares that a base model continues raw text (instead of the runner inferring it from the task name).

## Legacy removal

Since every task is now packaged, the following are gone (net −260 lines):

- The fluency special cases in `config.py` and the pairwise runner — the runner has a single data path via the task's dataset adapter.
- The monolithic `judge-arena-dataset` pattern-matching fallback in `download_hf`; downloads go only through task-declared, revision-pinned sources.
- The legacy prompt-preset fallbacks (`TASK_DEFAULT_PRESET`, the `fluency-` name check) — presets come from the task YAML.
- `dataset_revisions.py` deleted entirely; every revision pin lives in the task definition that uses it.

# Notes:
Entire stack is tested with running smoke tests for each of the defined task. No error is raised.


judgearena/benchmarks/pairwise/runner.py
```diff
@@ -40,39 +28,6 @@
 logger = get_logger(__name__)
 
 
-def _try_load_legacy_dataset_completions(
```
kargibora, line 43: As we have moved everything in the current pipeline as a task, we can get rid of these finally

geoalgo (commented):
Lines changed: 226 additions & 294 deletions
autoaccept, 9 stacked PRs are *almost* conquered!

geoalgo (approved):

PR #94, opened 2026-08-08
Cache: WIP (1/7) Add content-addressed local inference caches

ErlisLushtaku:
# Description

> [!NOTE]
> WIP: This cache stack is on hold until the task-YAML stack ending in #93 merges. The current implementation is not final and may change when it is adapted to the merged task runtime.

Adds content-addressed local SQLite caches for model completions and judge outputs.

Cache layout:

```text
{store_root}/completions/{task}/{provider}/{model}/{descriptor_hash}/
├── metadata.json
└── completions.db

{store_root}/judgements/{task}/{provider}/{model}/{descriptor_hash}/
├── metadata.json
└── judgements.db
```

`metadata.json` stores the validated model descriptor. The databases use these schemas:

```sql
completions(
  input_hash TEXT PRIMARY KEY,
  input_text TEXT NOT NULL,
  completion TEXT NOT NULL,
  benchmark TEXT NOT NULL,
  instruction_id TEXT NOT NULL,
  model TEXT NOT NULL,
  pushed_by TEXT NOT NULL,
  pushed_at TEXT NOT NULL,
  run_id TEXT NOT NULL
)

judgements(
  input_hash TEXT PRIMARY KEY,
  judge_input TEXT NOT NULL,
  judge_completion TEXT NOT NULL,
  benchmark TEXT NOT NULL,
  instruction_id TEXT NOT NULL,
  model_a TEXT NOT NULL,
  model_b TEXT NOT NULL,
  judge TEXT NOT NULL,
  orientation TEXT,
  pushed_by TEXT NOT NULL,
  pushed_at TEXT NOT NULL,
  run_id TEXT NOT NULL
)
```

This is stacked on #71.

PR #102, opened 2026-08-15
Meta-eval: (2/7) Score sampled battles with accuracy and Cohen's kappa

ErlisLushtaku:
# Description

> [!NOTE]
> Packaged meta-eval split from closed PR #76. Meta-eval does not use the unified `do_inference` cache, and the cache stack (#94-#100) does not wire it up either: caching will be added to meta-eval once those PRs are merged.

Judges the sampled battles in stored A/B order and reports accuracy and Cohen's kappa with bootstrap SE. Agreement is written for all battles and for the no-human-tie subset.

PairScore temperature is 0.5 here (paper / Figure 4); generate+judge stays 0.3.

This is stacked on #101.

### DIFF
--- a/judgearena/benchmarks/meta_eval/annotate.py
+++ b/judgearena/benchmarks/meta_eval/annotate.py
@@ -1,0 +1,105 @@
+"""Judge a sampled arena battle set in the stored A/B order."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import numpy as np
+import pandas as pd
+
+from judgearena.arenas_utils import extract_turn_text
+from judgearena.evaluate import PairScore, annotate_battles
+
+if TYPE_CHECKING:
+    from judgearena.config import RunConfig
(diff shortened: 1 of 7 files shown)


judgearena/benchmarks/meta_eval/annotate.py
```diff
@@ -0,0 +1,105 @@
...
+from judgearena.arenas_utils import extract_turn_text
+from judgearena.evaluate import PairScore, annotate_battles
+
+if TYPE_CHECKING:
+    from judgearena.config import RunConfig
+    from judgearena.prompts.registry import ResolvedJudgePrompt
+
+TIE_EPSILON = 0.01
+# Paper PairScore temperature for meta-eval; generate+judge keeps 0.3.
+META_EVAL_PAIRSCORE_TEMPERATURE = 0.5
```
ErlisLushtaku, line 19: q: should we align it with the other cases where we use `0.3`?

PR #105, opened 2026-08-15
Meta-eval: (5/7) Add Arena-Hard and AlpacaEval judge prompt presets

ErlisLushtaku:
# Description

> [!NOTE]
> Packaged meta-eval split from closed PR #76. Meta-eval does not use the unified `do_inference` cache, and the cache stack (#94-#100) does not wire it up either: caching will be added to meta-eval once those PRs are merged.

Meta-eval selects Arena-Hard and AlpacaEval prompts through `--judge.prompt_preset` rather than a separate `prompt_mode` flag. `arena-hard` parses Likert tags, `alpaca-eval` parses `ordered_models` JSON, and `alpaca-eval-pair-score` keeps the AlpacaEval prompt with PairScore output. The task default remains PairScore.

PairScore preferences exactly on the tie epsilon are classified as ties, matching the original meta-eval parser.

This is stacked on #104.

### DIFF
--- a/judgearena/config.py
+++ b/judgearena/config.py
@@ -17,6 +17,7 @@ from pydantic_settings import (
 )
 
 from judgearena.benchmarks.pairwise.baselines import native_pairwise_baseline
+from judgearena.prompts.registry import PRESETS
 from judgearena.tasks.registry import get_packaged_task
 from judgearena.tasks.schema import EloProtocol, MetaEvalProtocol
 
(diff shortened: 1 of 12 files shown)


judgearena/config.py
```diff
@@ -478,6 +479,17 @@ def _validate(self) -> RunConfig:
...
+        # parsers; elsewhere the judge output would only fail to parse after
+        # every judge call has been paid for.
+        preset = PRESETS.get(self.judge.prompt_preset or task_judge.default_prompt)
+        if preset is not None and preset.parser_mode != "score":
+            raise ValueError(
+                f"judge.prompt_preset={preset.name!r} produces "
+                f"{preset.parser_mode!r} verdicts, which only meta-eval tasks "
+                f"can parse; task {self.task!r} needs a score-based preset."
+            )
+
```
ErlisLushtaku, line 492: q: should we make these presets available to the other benchmarks as well?

PR #108, opened 2026-08-20
Official Task Variations (1/3): Refactor Prompt Parsing into Reusable Components

kargibora:
## Problem

Prompt presets previously returned a parser-mode string, while `evaluate.py` contained the parser implementation. Adding a judge output format required changes in both places. Custom prompt files also had to use the score parser, even when their output had another format.

Parsers also returned only one scalar preference. This discarded useful information such as the original A/B scores or a normalized verdict label.

## Changes

- Add `JudgeParser` implementations and a parser registry in `judgearena/prompts/parsing.py`.
- Make each prompt preset own the parser that matches its output format.
- Let custom prompt files select a named parser through `judge.prompt`.
- Add `ParsedPreference`, which contains a canonical preference and optional `label`, `scores`, and `details`.
- Keep existing scalar parser calls compatible through `parser(...)`.
- Store the structured parser result on `JudgeAnnotation.parsed`.
- Pass the resolved parser through pairwise, Elo, and MT-Bench judging.
- Keep swapped annotation values in the order shown to the judge. Only the scoring preference is converted to the model order used downstream.
- Move packaged prompt files under `judgearena/prompts/templates/` without changing their text.
- Accept only unused legacy `false`/`null` judge fields; reject active removed settings and unknown fields.

## Parser result

```python
ParsedPreference(
    preference=0.6457,
    scores={"A": 6.0, "B": 8.0},
)
```

`preference` uses the judge input order:

- `0.0`: A wins
- `0.5`: tie
- `1.0`: B wins
- intermediate values: continuous preferences
- `None` result: unparseable output

PairScore stores its raw A/B scores. Other parsers can store a normalized label, numeric values, or parser-specific details. This PR does not add per-criterion scoring.

## Flow

```text
Task YAML or run config
└── resolve prompt preset
    └── ResolvedJudgePrompt
        ├── system prompt
        ├── user prompt template
        └── JudgeParser
            └── ParsedPreference
                ├── retained on JudgeAnnotation
                └── preference projected into the scoring data
```

## Custom prompt example

```yaml
judge:
  model: OpenRouter/example/judge
  prompt:
    system_file: prompts/system.txt
    user_file: prompts/user.txt
    parser: score
```

Existing tasks keep their default prompt and PairScore behavior unless a task or run configuration selects another preset or parser.

## Tests

The tests cover prompt and parser resolution, custom prompt validation, structured PairScore output, scalar parser compatibility, parser handoff, and swapped preference handling.

```bash
uv run pytest -q tests
```

Result: `298 passed` locally.

### Raw diff

Relative to `main` at `df363e0`. Parent changes are excluded. Pre-existing tests are restored; remaining test reductions apply only to stack-added coverage.

|---|---:|---:|---:|
| Production Python | 595 | 390 | 985 |
| Tests | 343 | 97 | 440 |
| Prompts, YAML, docs, packaging | 34 | 22 | 56 |
| **Total** | **972** | **509** | **1,481** |

```
 ┌──────────────────────────────┬───────────────┐
 │ Type of change               │ Estimated LoC │
 ├──────────────────────────────┼───────────────┤
 │ New implementation           │ 380–430       │
 ├──────────────────────────────┼───────────────┤
 │ Moved or extracted code      │ 20–30         │
 ├──────────────────────────────┼───────────────┤
 │ Small refactors and wrappers │ 140–180       │
 ├──────────────────────────────┼───────────────┤
 │ Removed implementation       │ 180–210       │
 └──────────────────────────────┴───────────────┘
```
 Most of the new code covers structured parser results, logprob handling, category-specific prompts, random pair ordering, and canonical battle rows.

 The main removal is the old 148-line evaluate_completions function. Eight prompt files were renamed without content changes, so they add no implementation LoC.


judgearena/benchmarks/mt_bench/prompt_templates.py
```diff
@@ -12,7 +12,7 @@
 def load_mt_bench_prompt_text(filename: str) -> str:
     return (
         files(_PROMPT_PACKAGE)
-        .joinpath("mt_bench", filename)
```
kargibora, line 15: This is a bit optional but I thought we can move them to a better place, as now `judgearena/prompts` includes code about the prompt and parsing, not just the template

judgearena/benchmarks/pairwise/runner.py
```diff
@@ -28,6 +29,57 @@
 logger = get_logger(__name__)
 
 
+def _build_judge_batches(
```
kargibora, line 32: Note: **This is required** to allow **different categories/languages** to receive different **prompt**. Personally this will be useful if we want to, for example, use translated prompt for each different language. At the moment, it will be required by the arena-hard as `creative_writing` judge is different, as well as its user prompt.

judgearena/benchmarks/pairwise/runner.py
```diff
@@ -28,6 +29,57 @@
...
+        group_index = categories.index[categories == category]
+        preset = category_prompts.get(category)
+        prompt = (
+            resolved_prompt if preset is None else resolve_judge_prompt(preset=preset)
+        )
+        groups.append((prompt, group_index))
+    return groups
+
+
+def _random_swap_mask(instructions: pd.Series) -> pd.Series:
```
kargibora, line 70: This is deterministic and hence, `seed=0`. We can make it non-deterministic later

judgearena/models.py
```diff
@@ -649,6 +712,9 @@ def make_model(model: str, max_tokens: int | None = 8192, **engine_kwargs):
             seed=seed,
             top_k_via_model_kwargs=True,
         )
+        if top_logprobs is not None:
```
kargibora, line 721: This feels very specific implementation to me but I dont know how to achieve this without changing ChatVLLM. we should either let it stay general or impelemnt a custom workflow but it can stay (we may want to output a single score and take a logprob of it to report uncertainity)

This will be used by arena-hard

judgearena/benchmarks/elo/runner.py
```diff
@@ -325,11 +325,11 @@ def run_judge() -> pd.DataFrame:
             completions_A=completions_A,
             completions_B=completions_B,
             swap_mode=cfg.judge.swap_mode,
-            provide_explanation=cfg.judge.provide_explanation,
```
kargibora, line 328: After the current changes, `provide_explanation` is just a different prompt preset. No need to specify it here UNLESS we want something to be aded to prompt.

judgearena/benchmarks/pairwise/runner.py
```diff
@@ -171,28 +224,75 @@ def _load_or_generate_completions(model_spec: str, *, role: str) -> pd.Series:
...
+    prompt_groups = _build_judge_batches(
+        cfg, resolved_task, instructions_df, eval_index, resolved_prompt
     )
 
-    eval_instruction_index = instructions.head(n_instructions).index.tolist()
+    # The baseline fills the first judged slot unless the deterministic mask
+    # switches the pair. Preferences are re-oriented to the canonical
+    # model/baseline frame after parsing.
+    swap_mask = (
+        _random_swap_mask(instructions) if cfg.judge.swap_mode == "random" else None
```
kargibora, line 236: For now this is deterministic, can add seed laters.

judgearena/benchmarks/pairwise/scoring.py
```diff
@@ -3,26 +3,19 @@
 from __future__ import annotations
 
 from collections.abc import Callable
-from dataclasses import dataclass
 
 import pandas as pd
 
 from judgearena.utils.eval import PrefSummary, compute_pref_summary
 
+PairwiseScoreFn = Callable[[pd.DataFrame], PrefSummary]
```
kargibora, line 11: Over-engineered design. Kept it simple. Wecan change it later without changing anything else in downstream if we want.

judgearena/tasks/schema/baselines.py
```diff
@@ -18,14 +18,12 @@ class RuntimeRequiredBaseline(StrictFrozenModel):
 class TaskDefaultBaseline(StrictFrozenModel):
     strategy: Literal["task_default"]
     reference_id: str = Field(min_length=1)
-    allow_runtime_override: bool = True
```
kargibora, line 21: This have been discussed but removed this constraints. No need for rejecting override from the cfg.

judgearena/benchmarks/elo/runner.py
```diff
@@ -325,11 +325,11 @@ def run_judge() -> pd.DataFrame:
...
             strip_thinking_before_judging=cfg.judge.strip_thinking_before_judging,
             system_prompt=resolved_prompt.system_prompt,
             user_prompt_template=resolved_prompt.user_prompt_template,
             prompt_preset=resolved_prompt.preset_name,
+            parse=resolved_prompt.parser,
```
ErlisLushtaku, line 332: The resolved parser is used here, however soft ELO [reparses every stored completion with `PairScore`](https://github.com/OpenEuroLLM/JudgeArena/blob/303a54c31b18c179a0432cc14dcd52643d894c98/judgearena/benchmarks/elo/runner.py#L517-L541) later, so a verdict or custom parser is discarded and those rows can become unparseable.

judgearena/config.py
```diff
@@ -398,38 +426,28 @@ def _validate(self) -> RunConfig:
         task_judge = protocol.judge
         if "swap_mode" not in self.judge.model_fields_set:
             self.judge.swap_mode = task_judge.default_swap_mode
-        if self.judge.swap_mode not in task_judge.allowed_swap_modes:
```
ErlisLushtaku, line 401: `judge.swap_mode=random` is accepted for MT-Bench after this validation is removed, however both MT-Bench judging paths still support only `fixed` and `both`, so the run fails after generation. I think we should validate modes against the selected runner before starting the run.

judgearena/prompts/parsing.py
```diff
@@ -0,0 +1,111 @@
...
+
+    name: str
+    """Registry key and run-metadata identifier."""
+
+    requires_top_logprobs: bool = False
+    """Whether judging should collect first-token top logprobs for this
+    parser (the backend must also be asked for them via judge.top_logprobs)."""
+
+    @abc.abstractmethod
+    def __call__(
```
ErlisLushtaku, line 42: This contract can return only one preference, so the criteria use case in the description would still need another parser interface change before per-criterion scores can reach annotations or scorers. I think we should either return a parsed result with the preference and structured values now, or note that we are leaving the criteria template for a follow-up.

judgearena/benchmarks/pairwise/runner.py
```diff
@@ -205,8 +305,22 @@ def _load_or_generate_completions(model_spec: str, *, role: str) -> pd.Series:
 
     df.to_csv(res_folder / f"{name}-annotations.csv", index=False)
 
-    scorer = PAIRWISE_SCORERS[resolved_task.spec.protocol.scoring.adapter]
-    summary = scorer.summarize(prefs)
+    # Scorers see one canonically-oriented row per judged battle; under
+    # swap_mode="both" every instruction contributes two rows.
+    repeats = 2 if cfg.judge.swap_mode == "both" else 1
+    battles = pd.DataFrame(
```
ErlisLushtaku, line 311: The scorer input is wider now, however it is built from a fixed dict and drops task fields such as `category` or language, which are one of the reasons given for widening it. Could we start from the aligned `instructions_df` rows and add the model, baseline, completion and preference columns so task scorers receive their declared data?

judgearena/prompts/parsing.py
kargibora: Criteria will be updated later. I had implemented it in a different branch where I have returned more structure from the `Parser`. Currently criteria is not integrated into the pipeline in any places so this can be a easy next step for us :)

judgearena/benchmarks/pairwise/runner.py
kargibora: This is fixed in the #113. Did some design changes in between hence there are some gaps between this PR and next one for the scorings

judgearena/benchmarks/elo/runner.py
kargibora: This structured change a bit in #113, but I will resolve this. I agree, many prompts can provide soft label to us, not just scores.

judgearena/config.py
kargibora: `allowed_swap_modes` introduces extra complexity, we had some protection towards some parameters before.  We can just put this guard within mt-bench pipeline; and extend it if we see the need for it

judgearena/prompts/parsing.py
```diff
@@ -0,0 +1,158 @@
...
+import abc
+import math
+import re
+from dataclasses import dataclass, field
+
+from judgearena.utils import strip_thinking_tags
+
+
+@dataclass(slots=True)
+class ParsedPreference:
```
kargibora, line 14: After Parsing it, we create a `ParserPreference` object which collects all the important informations such as the preference value or scores.

judgearena/prompts/parsing.py
kargibora: Updated with the last commit, now we are returning more structurede values.

ErlisLushtaku (approved):
Looks good, just one comment regarding MT-Bench validation left

PR #113, opened 2026-09-03
Official Task Variations (2/3): Make Scoring more Composable

kargibora:
## Problem

Pairwise, Elo, and MT-Bench used separate scoring and reporting paths. Metric parameters were also split between task definitions and runner code. Adding a metric could therefore require changes in several pipelines.

Elo also handled fitting, bootstrap aggregation, calibration, and rendering inside its runner. This made the scoring boundary difficult to reuse.

## Changes

- Add configured, stateless metric classes with `calculate(battles)` and `render(result)` methods.
- Add the shared metric functions `available_metrics()`, `build_metric()`, `build_metrics()`, `calculate_metrics()`, and `render_metrics()`.
- Add declarative `MetricSpec` and `ScoringSpec` task configuration.
- Pass metric parameters through `MetricSpec.parameters`.
- Use one scoring boundary: battle dataframe to metric calculation to plain result dictionary.
- Use the same metric construction and execution path for pairwise, Elo, and MT-Bench.
- Move report models into `judgearena/reports.py`.
- Move PairScore temperature calibration into `judgearena/benchmarks/elo/calibration.py`.
- Make Bradley-Terry own fitting, bootstrap ratings, human-reference ratings, uncertainty, and MAE.
- Preserve aligned instruction fields, such as category and language, when pairwise runners build battle rows.
- Build MT-Bench rows with explicit model names and direct or reversed orientation.

## Parsed preferences and Soft Elo

Soft Elo uses the continuous preference returned by the configured parser. Verdict, token-logprob, and custom parser preferences are not replaced with PairScore parsing.

Temperature adjustment and calibration remain specific to `PairScore`. PairScore calibration reads the raw A/B scores stored in `JudgeAnnotation.parsed`.

The scoring path is:

```text
JudgeParser
└── ParsedPreference
    └── canonical numeric pref
        └── battle dataframe
            └── configured metrics
                └── plain result dictionaries
```

Metrics consume the numeric `pref` column. Structured parser evidence remains on the judge annotation and can be added to battle rows later when a metric has a defined use for it.

## Task configuration

Metrics and their parameters are selected in task YAML:

```yaml
scoring:
  metrics:
    - metric: pairwise_win_rate
      breakdown_by: [category]
    - metric: bradley_terry
      parameters:
        n_bootstraps: 0
        soft: false
```

Runners use the same execution path:

```python
metrics = build_metrics(protocol.scoring.metrics)
results = calculate_metrics(battles, metrics)
```

Runtime-only values, such as a random number generator, remain explicit and are not serialized as metric configuration.

A new metric needs a configured metric class, its calculation and rendering methods, one internal registry entry, and a task YAML entry. It does not need a separate scoring implementation in each runner.

## Tests

The tests cover shared metric construction, configuration validation, pairwise and MT-Bench battle rows, hard and soft Bradley-Terry behavior, PairScore calibration, generic continuous parser preferences, report rendering, and runtime dependencies.

```bash
uv run pytest -q tests
```

Result: `328 passed` locally.

### Raw diff

Relative to #108 at `95b7e12`. Parent changes are excluded. Pre-existing tests are restored; remaining test reductions apply only to stack-added coverage.

|---|---:|---:|---:|
| Production Python | 1,202 | 568 | 1,770 |
| Tests | 1,019 | 109 | 1,128 |
| YAML and documentation | 33 | 10 | 43 |
| **Total** | **2,254** | **687** | **2,941** |

```
 ┌──────────────────────────────┬───────────────┐
 │ Type of change               │ Estimated LoC │
 ├──────────────────────────────┼───────────────┤
 │ New implementation           │ 420–470       │
 ├──────────────────────────────┼───────────────┤
 │ Moved or extracted code      │ 270–320       │
 ├──────────────────────────────┼───────────────┤
 │ Small refactors and wrappers │ 150–190       │
 ├──────────────────────────────┼───────────────┤
 │ Removed implementation       │ 20–40         │
 └──────────────────────────────┴───────────────┘
```
 The new implementation is mainly the composable metric framework and length-controlled win-rate calculation.

 Most of the movement comes from taking calibration, Bradley–Terry calculation, and report rendering out of the Elo runner. These were moved into dedicated metric, calibration, and report modules. The runner changes are therefore mostly extraction and integration rather than new calculation code.


judgearena/benchmarks/elo/calibration.py
```diff
@@ -0,0 +1,143 @@
+"""PairScore temperature calibration against human arena preferences."""
```
kargibora, line 1: This script should be self-contained. It was sitting on the `runner.py` which was making it harder to read; so its a simple refactor and not a new capability. It's only purpose is to use some rows to calculate a temperature.

judgearena/benchmarks/elo/rating.py
```diff
@@ -142,7 +142,9 @@ def prefs_to_battle_results(
     for pref, is_pos_a, opponent in zip(
         prefs, our_model_is_position_a, opponent_models, strict=True
     ):
-        if _is_nan_pref(pref) or pref == 0.5:
+        if _is_nan_pref(pref):
```
kargibora, line 145: Making nan pref's tie does not make sense. Tie is basically a indicator that both models perform same. ELO calculation for example makes these model ELO's closer to each other if they are tie. Thus it should be marked as NaN correctly.

judgearena/benchmarks/elo/runner.py
```diff
@@ -13,130 +13,36 @@
...
-from judgearena.utils.eval import PrefSummary, Report
+from judgearena.utils import cache_function_dataframe
 
 if TYPE_CHECKING:
     from judgearena.config import RunConfig
 
 logger = get_logger(__name__)
 
 
-class EloReport(Report):
```
kargibora, line 47: Moved into `report.py`. Metric reporting is now contained within the metric, so they are re-usable. `Report` is just collection of `key:values` which we want to add to the result.

judgearena/benchmarks/elo/runner.py
```diff
@@ -395,124 +300,26 @@ def run_judge() -> pd.DataFrame:
...
-            for p, is_pos_a in zip(prefs, our_model_is_position_a, strict=True)
-        ]
-    )
-    summary = compute_pref_summary(prefs_normalized)
-
     # Anchor the llm-judge battles against the human arena battles. These are
     # rebuilt from the (revision-pinned) arena, not persisted per run.
     df_arena = arena_anchor_battles(df_arena_all)
 
-    df_results = pd.concat([df_llm_judge, df_arena], ignore_index=True)
```
kargibora, line 422: No need for all the `calibration` code within the runner.

judgearena/benchmarks/elo/runner.py
```diff
@@ -540,83 +347,65 @@ def run_judge() -> pd.DataFrame:
...
-    method_label = "Soft-ELO" if use_soft else "ELO"
-
-    # Count battles per model across the combined results
-    battle_counts: dict[str, int] = {}
-    for _, row in df_results.iterrows():
-        battle_counts[row["model_a"]] = battle_counts.get(row["model_a"], 0) + 1
-        battle_counts[row["model_b"]] = battle_counts.get(row["model_b"], 0) + 1
-
-    pref_col = "pref" if use_soft else "pref_hard"
-    bootstrap_ratings: list[dict[str, float]] = []
```
kargibora, line 568: These are `metrics`. Not related to the `runner.py`. Thus they are moved into the classes which is going to use, in this case, `BradleyTerryMetric` which is basically ELO.

judgearena/benchmarks/elo/runner.py
```diff
@@ -540,83 +347,65 @@ def run_judge() -> pd.DataFrame:
...
+    # anchors keep these columns null. Focal metrics can therefore consume the
+    # same combined battle table as Bradley-Terry without knowing this runner.
+    repeats = max(1, len(df_llm_judge) // max(1, len(our_completions)))
+    row_our_completions = (list(our_completions) * repeats)[: len(df_llm_judge)]
+    row_opponent_completions = (list(opponent_completions) * repeats)[
+        : len(df_llm_judge)
     ]
-    elo_mean = (
-        float(np.mean(model_rating_values)) if model_rating_values else float("nan")
+    focal_is_a = pd.Series(our_model_is_position_a, dtype="bool")
```
kargibora, line 368: Normalizes some column. I want to make the boundaries more clear so datasets are interchangable at the core (basically arena and pairwise pipelines almost same. The dataset includes a prompt and two responses, in arena it is between each model and in pairwise, it is between some baseline and other models. For now as we already normalized ELO datasets, this will allow us to use pairwise metrics easily (lenght contorlled etc).

judgearena/benchmarks/elo/runner.py
```diff
@@ -540,83 +347,65 @@ def run_judge() -> pd.DataFrame:
...
+            "bradley_terry": {
+                "n_bootstraps": cfg.elo.n_bootstraps,
+                "baseline_model": cfg.elo.baseline_model,
+                "soft": cfg.elo.soft_elo,
+            },
+        },
     )
-    elo_std = (
-        float(np.std(model_rating_values)) if model_rating_values else float("nan")
+    metric_results = calculate_metrics(
```
kargibora, line 395: Pay-off. This will handle any metrics. No need to change anything in the pipeline when we add newer metric.

judgearena/benchmarks/elo/scoring.py
```diff
@@ -1,28 +1,207 @@
...
+
+@dataclass(frozen=True, kw_only=True)
+class BradleyTerryMetric:
+    """Configured arena-anchored Bradley-Terry calculation."""
+
+    n_bootstraps: int = 0
+    baseline_model: str | None = None
+    soft: bool = True
+
+    def __post_init__(self) -> None:
```
kargibora, line 22: This allow us to construct the class without setting some default values in the init.

judgearena/benchmarks/pairwise/scoring/metrics.py
```diff
@@ -0,0 +1,287 @@
...
+
+def _has_separation(x: np.ndarray, outcomes: np.ndarray) -> bool:
+    positive = x[outcomes > 0]
+    negative = x[outcomes < 1]
+    if not len(positive) or not len(negative):
+        return True
+    return bool(positive.max() <= negative.min() or negative.max() <= positive.min())
+
+
+def _fit_length_model(
```
kargibora, line 165: Estimate the effect of the length differences in predicting the winrate.

judgearena/benchmarks/pairwise/runner.py
```diff
@@ -102,13 +102,36 @@ def run_pairwise(cfg: "RunConfig", resolved_task: ResolvedTaskSpec | None = None
...
+        "completion_baseline",
+        "pref",
+        "orientation",
+        "judge",
+        "judge_prompt_preset",
+        "judge_temperature",
+        "judge_max_out_tokens",
+        *instructions_df.columns,
+    }
+    missing_groups = sorted(
```
kargibora, line 133: Allow us to group_by orientation for example (shows normal and swapped winrates), or any category/lang exist in the task (allow us to also report winrates per lang easily).

judgearena/utils/eval.py
```diff
@@ -44,141 +42,10 @@ def compute_pref_summary(prefs: pd.Series) -> PrefSummary:
     )
 
 
-class Report(BaseModel, abc.ABC):
```
kargibora, line 47: Moved into `report.py`

judgearena/reports.py
```diff
@@ -0,0 +1,90 @@
...
+        arbitrary_types_allowed=True,
+        populate_by_name=True,
+        protected_namespaces=(),
+        use_attribute_docstrings=True,
+    )
+
+    @computed_field
+    @property
+    def schema_version(self) -> str:
+        return "1"
```
ErlisLushtaku, line 25: The saved result layout changes from flattened summary fields to a nested `metrics` object, however `schema_version` stays at `1`. Should we bump this?

judgearena/tasks/schema/metrics.py
```diff
@@ -0,0 +1,36 @@
...
+from pydantic import Field, model_validator
+
+from judgearena.tasks.schema.base import StrictFrozenModel
+
+
+class MetricSpec(StrictFrozenModel):
+    """One named calculation over a battle dataframe."""
+
+    metric: str = Field(min_length=1)
+    group_by: tuple[str, ...] = ()
```
ErlisLushtaku: `group_by: [category, turn]` normally means one grouping by the `(category, turn)` unique combination, however [the executor](https://github.com/OpenEuroLLM/JudgeArena/blob/6bb0a652a8de0a49112e21bbc5efef9284b9dfab/judgearena/benchmarks/scoring.py#L103) calculates a separate breakdown for each field. Should we rename this to something like `breakdown_by` so isn't confused?

judgearena/reports.py
kargibora: I think we can leave it out as it is. Current versioning is hard to track as we are casually improving architecture. This is meaningful when our repo is stable and we introduce things slowly (imo)

ErlisLushtaku: Makes sense

ErlisLushtaku (approved):
Just a small concern regarding the semantics of group_by.

PR #115, opened 2026-09-03
Official Task Variations (3/3): Add Official AlpacaEval, Arena-Hard, and MT-Bench Tasks

kargibora:
This PR is stacked on #113. It adds the official AlpacaEval, Arena-Hard, and MT-Bench task protocols.

## Problem

The existing task names used JudgeArena defaults rather than the original benchmark protocols. This made it difficult to reproduce official results or compare a JudgeArena run with the upstream benchmark.

## Canonical tasks

The canonical task names now use the official defaults:

| Task | Default protocol |
|---|---|
| `alpaca-eval` | Pinned AlpacaEval 2.0 data, `gpt4_1106_preview` baseline, official prompt, random answer order, first-token logprobs, and official length-controlled scoring |
| `arena-hard-v0.1` | Official v0.1 data and baseline, official verdict prompt, both answer orders, and official weighted scoring |
| `arena-hard-v2.0` | Official v2.0 data, category-specific baselines and prompts, both answer orders, and judge-specific style-controlled scoring |
| `mt-bench` | Pinned FastChat questions and references, `gpt-3.5-turbo` baseline, FastChat pairwise prompt, and both answer orders |

Normal runtime options can still override model and generation settings.

## JudgeArena variants

The existing JudgeArena protocols remain available under explicit names:

| Task | Protocol |
|---|---|
| `alpaca-eval-ja` | JudgeArena prompt, fixed answer order, pairwise win rate, and JudgeArena length-controlled win rate |
| `arena-hard-v0.1-ja` | JudgeArena prompt, fixed answer order, and pairwise win rate |
| `arena-hard-v2.0-ja` | Official v2.0 data and category baselines with the JudgeArena prompt, fixed answer order, and pairwise win rate |

These are separate protocols rather than aliases for the canonical tasks.

## Implementation

- Add the official AlpacaEval length-controlled metric.
- Add the official Arena-Hard v0.1 weighted metric.
- Add the official Arena-Hard v2 style-controlled metric and packaged calibration data.
- Put data sources, revisions, baselines, prompts, generation settings, and metric parameters in task YAML.
- Reuse the existing JudgeArena table adapter for AlpacaEval.
- Keep the native Arena-Hard adapter for its versioned JSONL and nested answer formats.
- Use the shared battle-dataframe and metric interfaces from #113.
- Route MT-Bench through its FastChat-compatible prompt and judging path.
- Require an OpenRouter provider that supports top logprobs when a task requests them.

The Arena-Hard sources contain:

- v0.1: 500 instructions, 37,510 outputs, and 76 models
- v2.0: 750 instructions, 22,496 outputs, and 30 models

## Parsed judge output

The official parsers use the structured parser result introduced in #108:

- AlpacaEval stores the normalized preference, emitted `m` or `M` label, and the relevant raw token logprobs.
- Arena-Hard stores the graded preference and final normalized verdict, such as `A>B` or `A>>B`.

Metrics continue to receive the canonical numeric `pref` column.

## Protocol constraints

AlpacaEval requires first-token top logprobs. Responses without them remain unparsed; textual labels are not used as a fallback. If every preference is missing, the metric is saved as unavailable.

Arena-Hard v2 style control is judge-specific. The packaged calibration supports `gpt-4.1` and `gemini-2.5` with their required settings. Other judges need their own calibration data. `arena-hard-v2.0-ja` remains available for pairwise evaluation with arbitrary judges.

## Tests

The tests cover task configuration, dataset normalization, prompt and parser behavior, official numerical results, Arena-Hard calibration, MT-Bench integration, OpenRouter capability routing, and package contents.

```bash
uv run pytest -q tests
```

Result: `359 passed` locally.

### Raw diff

Relative to #113 at `6a42472`. Parent changes are excluded. Pre-existing tests are restored; remaining test reductions apply only to stack-added coverage.

|---|---:|---:|---:|
| Production Python | 1,284 | 128 | 1,412 |
| Tests | 804 | 125 | 929 |
| Task YAML, prompts, docs, packaging | 237 | 45 | 282 |
| **Total** | **2,325** | **298** | **2,623** |

```
 ┌─────────────────────────┬───────────────┐
 │ Type of change          │ Estimated LoC │
 ├─────────────────────────┼───────────────┤
 │ New implementation      │ 1,095–1,120   │
 ├─────────────────────────┼───────────────┤
 │ Moved or extracted code │ 45–70         │
 ├─────────────────────────┼───────────────┤
 │ Small refactors         │ 5–10          │
 ├─────────────────────────┼───────────────┤
 │ Removed implementation  │ 55–65         │
 └─────────────────────────┴───────────────┘
```


judgearena/benchmarks/mt_bench/fastchat_compat.py
```diff
@@ -26,18 +26,6 @@
 from judgearena.prompts.registry import DEFAULT_JUDGE_PROMPT_PRESET
 from judgearena.utils import strip_thinking_tags
 
-FASTCHAT_TEMPERATURE_CONFIG: dict[str, float] = {
```
kargibora, line 29: Moved it into the task definition

judgearena/datasets/pairwise.py
```diff
@@ -36,8 +36,13 @@ def __post_init__(self) -> None:
                 f"Pairwise model outputs are missing canonical columns: {missing}."
             )
 
-    def model_completion(self, model: str) -> pd.Series | None:
-        """Return model completions aligned to the instruction index."""
+    def model_completion(
+        self,
+        model: str,
+        *,
+        instruction_ids: pd.Index | None = None,
```
kargibora, line 43: We need `instruction_ids` because some official protocol asks for filtered instructions and groups to judge different prompts with different configurations

judgearena/evaluate.py
```diff
@@ -281,14 +281,12 @@ def _none_to_nan(x):
...
-            n_no_logprobs = sum(1 for a in ann_list if a.judge_top_logprobs is None)
+            n_no_logprobs = sum(1 for a in ann_list if not a.judge_top_logprobs)
             if n_no_logprobs:
-                logger.warning(
-                    "%d/%d judge responses returned no logprobs (%s) — falling "
-                    "back to discrete token parsing for those.",
-                    n_no_logprobs,
-                    len(ann_list),
-                    label,
+                raise ValueError(
```
kargibora, line 288: Should we skip or give warning or raise a `ValueError` here?

judgearena/models.py
```diff
@@ -707,6 +710,12 @@ def make_model(model: str, max_tokens: int | None = 8192, **engine_kwargs):
         # advertise the OpenRouter-native max_tokens parameter instead.
         extra_body = dict(openai_kwargs.pop("extra_body", {}) or {})
         extra_body["max_tokens"] = openai_kwargs.pop("max_tokens")
+        if top_logprobs is not None:
+            # OpenRouter may otherwise route individual requests through
+            # providers that silently omit requested logprobs.
+            provider = dict(extra_body.get("provider", {}) or {})
+            provider["require_parameters"] = True
```
kargibora, line 717: without this, some provider does not return `logprobs` and hence, these backends can be used during API routing which fails the pipeline

judgearena/benchmarks/pairwise/scoring/arena_hard_v20_calibration.csv.gz
ErlisLushtaku, line 1: Is this intentional?

judgearena/config.py
```diff
@@ -420,6 +420,18 @@ def _validate(self) -> RunConfig:
             )
 
         protocol = resolved_task.spec.protocol
+        task_generation = getattr(protocol, "generation", None)
```
ErlisLushtaku, line 423: The task defaults added here leave `generation.truncate_all_input_chars` at the generic 8192. The pinned Arena-Hard data has 2 v0.1 prompts and 15 v2.0 prompts over that limit, up to 27,586 chars, while upstream passes the full question to generation. Could official tasks set this default to `None`? Otherwise newly generated candidate answers use truncated prompts on those rows.

judgearena/benchmarks/pairwise/scoring/arena_hard.py
```diff
@@ -0,0 +1,525 @@
...
+class ArenaHardV01Metric:
+    """Official Arena-Hard v0.1 weighted score and uncertainty."""
+
+    def calculate(self, battles: pd.DataFrame) -> dict[str, object]:
+        _, ci_low, ci_high = _confidence_interval(
+            battles,
+            lower_quantile=V01_CI_LOWER_QUANTILE,
+            upper_quantile=V01_CI_UPPER_QUANTILE,
+            seed=42,
+            require_complete_pairs=False,
```
ErlisLushtaku, line 463: Upstream drops the whole judgment when either of the two games is missing [here](https://github.com/lmarena/arena-hard-auto/blob/196f6b826783b3da7310e361a805fa36f0be83f3/show_result.py#L30-L33), while `require_complete_pairs=False` keeps the parseable orientation. I think v0.1 should require complete pairs as well, otherwise one-sided parse failures change the official score.

judgearena/benchmarks/pairwise/scoring/arena_hard_v20_calibration.csv.gz
kargibora: Yes, arena hard uses some pre-defined CSV file to configure their scoring. Perhaps we can avoid having this `.csv` but should put these data somehow if we want true replication

judgearena/config.py
kargibora: Yes this is fair. I will fix it

judgearena/benchmarks/pairwise/scoring/arena_hard.py
kargibora: I will take a look at v0.1 but thanks for catching this

judgearena/benchmarks/pairwise/scoring/arena_hard_v20_calibration.csv.gz
ErlisLushtaku: Can this be regenerated or is it fixed?

judgearena/benchmarks/pairwise/scoring/arena_hard.py
kargibora: It seems 2024 version of Arena-Hard retains the other verdict while newer version (2025) drops the question if a verdict is missing. What to do here?

ErlisLushtaku: I see, then I think we should leave this as is so each version follows its corresponding upstream implementation.

PR #117, opened 2026-09-07
[feat]: Adding Meta-Eval pipeline into the framework

kargibora:
## Problem

   JudgeArena can use judges to evaluate models, but it needs a way to evaluate the judges themselves against human preferences. This PR adds that comparison within the existing task and metric framework. 

   ## What this PR introduces

   Meta-evaluation tasks for **LMArena-100k, LMArena-140k, and ComparIA**, with three types of comparison:

   - **Agreement:** accuracy and Cohen’s kappa against human votes, with parse coverage reported separately.
   - **Ranking:** Spearman correlation and Elo MAE between human and judge rankings.
   - **Elo-gap:** held-out Elo error at different annotation budgets, using hard and soft preferences.

All of these metrics can be improved,removed or added easily - for example one can easily add Kendells-$\tau$ to the correlation without changing the pipeline.

   ## How it works

   The runner selects the most frequent models and samples a connected set of battles with a per-model quota. It judges those battles through the existing prompt and parsing pipeline, then passes canonical numeric preferences to the configured metrics. 

   The implementation reuses the existing dataset loaders, `ParsedPreference`, Bradley–Terry fitter, and metric/report APIs. Task YAML defines the dataset, prompt, and metric settings. Runtime options control sampling.

   Both fixed and two-orientation judging are supported. With two orientations, both passes must parse before a battle contributes a preference. Failed judgments count against attempted agreement accuracy rather than disappearing from the denominator.
   For example:

   ```bash
   uv run judgearena \
     --task meta-eval-lmarena-140k-en \
     --judge.temperature 0 \
     --judge.swap_mode fixed \
     --meta_eval.top_models 10 \
     --run.seed 0 \
     --run.result_folder results/meta-eval
   ```

   The run saves the sampled battles, judge annotations, metric inputs, and results for inspection.

## Size and validation

Relative to #115 at `e38ccae`. Parent changes are excluded. Pre-existing tests are restored; remaining test reductions apply only to stack-added coverage.

|---|---:|---:|---:|
| Production Python | 1,486 | 77 | 1,563 |
| Tests | 601 | 6 | 607 |
| Task YAML, prompts, README | 199 | 1 | 200 |
| **Total** | **2,286** | **84** | **2,370** |

Most core additions are in scoring (576 lines), the runner (217), annotation (184), and sampling (164).

Full local suite: `389 passed`.

### DIFF
--- a/judgearena/benchmarks/elo/runner.py
+++ b/judgearena/benchmarks/elo/runner.py
@@ -14,6 +14,7 @@ from judgearena.artifacts import (
     write_run_metadata_safely,
 )
 from judgearena.battles import Leaderboard, RatingEntry, write_battles
+from judgearena.benchmarks.arena import resolve_task_languages
 from judgearena.benchmarks.elo.calibration import calibrate_pairscore_temperature
 from judgearena.benchmarks.elo.rating import (
     arena_anchor_battles,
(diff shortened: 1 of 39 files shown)


judgearena/benchmarks/elo/runner.py
```diff
@@ -58,22 +59,10 @@ def run_elo(cfg: "RunConfig", task: ResolvedTaskSpec | None = None) -> dict:
...
-            ]
-            if not selected_languages:
-                raise ValueError(
-                    f"elo.languages {cfg.elo.languages} has no overlap with the "
-                    f"languages of task {cfg.task!r} ({variant_languages})."
-                )
-        else:
-            selected_languages = variant_languages
+    # A task variant preselects languages; elo.languages may narrow it further.
+    selected_languages = resolve_task_languages(
```
kargibora, line 63: This is now re-used with the `elo`. No need for duplication

ErlisLushtaku (approved):

PR #22, opened 2026-03-09
feat: add support for mt-bench-101

ErlisLushtaku:
## Summary

Integrates the MT-Bench-101 benchmark (1,388 multi-turn dialogues, 13 tasks, 3 ability tiers) into OpenJury. The implementation follows the [OpenCompass fork](https://github.com/xingyuanbu/opencompass) reference implementation (as well as some info from https://github.com/mtbench101/mt-bench-101) and the [paper (Bai et al., 2024)](https://arxiv.org/pdf/2402.14762) for prompt structure, dialogue formatting, and scoring, while adding OpenJury-specific derived pairwise comparisons on top.

## What is MT-Bench-101

MT-Bench-101 is a fine-grained multi-turn dialogue evaluation benchmark that tests 13 specific abilities (e.g. Conversational Memory, Self-Correction, Mathematical Reasoning) grouped into three tiers: Perceptivity, Adaptability, and Interactivity. Unlike the original MT-Bench (80 dialogues, 2 turns, pairwise judging), MT-Bench-101 uses:

- **Golden context**: dialogue history is curated and fixed; the model generates only the current turn's response
- **Single-answer grading**: each model is scored independently (1-10) with task-specific judge prompts
- **Min-per-dialogue aggregation**: a dialogue's score is the minimum across its turns, then averaged per task/ability

## Key design decisions

- **Paper-faithful evaluation**: single-answer grading with judge temperature 0.6, task-specific prompts from the paper (Figures 21-34), and min-per-dialogue aggregation -- preserving the methodology that achieved 87% agreement with human experts
- **Derived pairwise on top**: after independent scoring, `PairScore.preference_from_scores` (softmax) derives pairwise preferences compatible with OpenJury's existing win-rate reporting
- **Aligned with OpenCompass**: judge prompt structure, dialogue formatting, and score parsing (`[[N]]` double-bracket extraction with 1-10 validation), prompt content (with some fixed typos) have been made consistent with the OpenCompass implementation.
- **Reference answer bug fix**: OpenCompass passes the full dialogue list repr as `ref_answer` for MR/GR tasks causing a bug; we correctly pass the ground-truth `bot` response for the current turn
[...]

PR #119, opened 2026-09-09
MT-Bench-101 (1/2): Declare MT-Bench-101 task and dataset adapter

ErlisLushtaku:
## Summary

This is the first PR in the MT-Bench-101 stack. It declares the task and adds the dataset adapter; generation and judging are in #120.

## Changes

- Add a packaged `mt-bench-101` task YAML with the upstream dataset pinned to commit `bc18b3e2c18c99164e11528f1a79c92083db5953`.
- Expand the source dialogues into one row per evaluated turn while retaining the golden context. Tasks that only become meaningful after the first turn start at turn 2, following the reference implementation.
- Add the typed protocol schema and register the dataset adapter and runner route.
- Package the task-specific judge prompts used by the OpenCompass fork, with the typo and formatting corrections carried over from closed #22.

## References

The task structure and prompts come from the [MT-Bench-101 repository](https://github.com/mtbench101/mt-bench-101), the [OpenCompass fork](https://github.com/xingyuanbu/opencompass), and [Bai et al. (2024)](https://arxiv.org/abs/2402.14762).


judgearena/tasks/schema/__init__.py
```diff
@@ -22,6 +22,11 @@
     MTBenchProtocol,
     MultiTurnGeneration,
 )
+from judgearena.tasks.schema.mt_bench_101 import (
+    GoldenContextGeneration,
```
kargibora, line 26: What is GoldenContext?

kargibora:
Looks good to me as it is just the definitions. IT is a clean implementation!

kargibora (approved):

judgearena/tasks/schema/__init__.py
ErlisLushtaku: It means each turn is generated using the preceding reference dialogue as fixed context, rather than using the previously generated turns from the LLM. I added this to the schema.

judgearena/tasks/schema/mt_bench_101.py
```diff
@@ -13,6 +13,8 @@
 
 
 class GoldenContextGeneration(StrictFrozenModel):
+    """Generate each turn using the preceding reference dialogue as context."""
```
kargibora, line 16: Thanks for clarifiyng!

kargibora (approved):

PR #122, opened 2026-09-09
Cache (2/5): Cache inference at do_inference

ErlisLushtaku:
# Description

Moves cache lookup and writes to `do_inference`.

- `PreparedModel` defers backend construction until an input misses the cache.
- Cache identity combines the canonical model input with the validated model descriptor.
- Mixed batches run only missing inputs and restore the original output order.
- Completion and judgement cache classes define their own stored rows.
- Cached judgements restore first-token logprobs. Hits do not record provider usage.

This is stacked on #121.


judgearena/inference.py
```diff
@@ -0,0 +1,210 @@
...
+    _model: Any = field(default=None, init=False, repr=False)
+
+    def materialize(self) -> Any:
+        if self._model is None:
+            self._model = self.factory()
+        return self._model
+
+
+@dataclass(frozen=True)
+class InferenceCache(ABC):
```
kargibora: Should we create a seperate file for storing these cache objects?

judgearena/models.py
```diff
@@ -772,6 +779,78 @@ def batch_with_retry(batch_inputs, max_retries=5, base_delay=1.0):
     return [result.text for result in results]
 
 
+def do_inference(
+    chat_model,
+    inputs,
+    use_tqdm: bool = False,
+    return_top_logprobs: bool = False,
+    *,
+    stage: str = "unspecified",
```
kargibora: Why it is unspecified?

judgearena/models.py
```diff
@@ -772,6 +779,78 @@ def batch_with_retry(batch_inputs, max_retries=5, base_delay=1.0):
...
+    cache = chat_model.cache
+    if cache is None or chat_model.descriptor is None:
+        return _do_inference_uncached(
+            chat_model.materialize(),
+            inputs,
+            use_tqdm,
+            return_top_logprobs,
+            stage=stage,
+        )
+    if cache_metadata is None or len(cache_metadata) != len(inputs):
```
kargibora: If `cache_metadata` is none or lengths fails we abort. But  then we check for `missing_indices` later. These are two different policies.

judgearena/models.py
```diff
@@ -772,6 +779,78 @@ def batch_with_retry(batch_inputs, max_retries=5, base_delay=1.0):
...
+        raise ValueError("cache_metadata must contain one row per inference input.")
+
+    input_texts = [canonicalize_chat_input(item) for item in inputs]
+    input_hashes = [input_hash(input_text) for input_text in input_texts]
+    with cache.open_store(chat_model) as store:
+        cached_rows = store.query(input_hashes).set_index("input_hash")
+        results: list[InferenceResult | None] = [
+            (
+                InferenceResult(**cache.cached_result(cached_rows.loc[key]).__dict__)
```
kargibora: Why we are converting `CachedInferenceResult` back to `InferenceResult`? If so we may not need `CachedInferenceResult`at all

judgearena/models.py
```diff
@@ -772,6 +779,78 @@ def batch_with_retry(batch_inputs, max_retries=5, base_delay=1.0):
...
+            generated = _do_inference_uncached(
+                chat_model.materialize(),
+                [inputs[index] for index in missing_indices],
+                use_tqdm,
+                True,
+                stage=stage,
+            )
+            for index, result in zip(missing_indices, generated, strict=True):
+                results[index] = result
+            cache.save_outputs(
```
kargibora: Can this fail? If it fails for some reason the entire pipeline fails. We can put try except

```
  try:
       cache.save_outputs(...)
   except CacheWriteError as exc:
       logger.warning("Could not save inference cache: %s", exc)
```

judgearena/models.py
```diff
@@ -772,6 +779,78 @@ def batch_with_retry(batch_inputs, max_retries=5, base_delay=1.0):
...
 
 
+def do_inference(
+    chat_model,
+    inputs,
+    use_tqdm: bool = False,
+    return_top_logprobs: bool = False,
+    *,
+    stage: str = "unspecified",
+    cache_metadata: list[dict] | None = None,
```
kargibora: I think one thing that is confusing for me is this PR introduces some changes like `cache_metadata` which will be provided in the future PR's in #124. Design-wise it makes sense to first introduce it without wiring however while reading the PR it creates some readability problems. I think if necessary, creating large PR's (~1000-2000LOC) where we introduce the functionality while introducing the changes makes more sense to me.

Not a requirement for this PR (as I will review the entire stack as a single one)

judgearena/models.py
```diff
@@ -663,7 +670,7 @@ def batch_inference_once(
     return [result.text for result in results]
 
 
-def do_inference(
+def _do_inference_uncached(
```
kargibora: We can rename this to `_call_model` or `_invoke_model` or similar (I am not creative enough). Because we will be using it to do inference for uncached samples, this reads like we will be doing inference without using the cache (although it is true we are doing it for uncached inputs, flow can be better)

judgearena/inference.py
ErlisLushtaku: Agreed, the cache-specific types now live in `judgearena/cache/inference.py` in #127. The provider-neutral `InferenceResult` stays in `judgearena/inference.py`.

judgearena/models.py
ErlisLushtaku: Yes, `CachedInferenceResult` is removed and cache hits now return `InferenceResult` directly.

judgearena/models.py
ErlisLushtaku: Good catch, cache write failures now log and preserve the generated results.

judgearena/models.py
ErlisLushtaku: Renamed it to `_run_backend_inference`.

judgearena/models.py
```diff
@@ -801,23 +797,53 @@ def do_inference(
...
+        )
+        return _run_backend_inference(
+            chat_model.materialize(),
+            inputs,
+            use_tqdm,
+            return_top_logprobs,
+            stage=stage,
+        )
+
+    try:
```
kargibora: I know we need this for closing the cache but it becomes hard to parse with lots of `try: excepts`

ErlisLushtaku: Agreed, I moved cache open and close handling into a context manager and read and write failures into small helpers.

kargibora (approved):
LGTM

PR #123, opened 2026-09-09
Cache (3/5): Cache inference across model providers

ErlisLushtaku:
# Description

Extends cache descriptors and input hashing across the providers exposed by `make_model`.

- Covers VLLM, OpenRouter, Dummy, ChatOpenAI, OpenAI, Together, and LlamaCpp.
- Records whether the backend receives chat messages or flattened text.
- Includes hosted endpoints, OpenRouter routing when present in `extra_body`, and output-affecting model settings.
- Excludes VLLM execution-only settings (`tensor_parallel_size`, `gpu_memory_utilization`, `enforce_eager`).
- Unsupported providers stay uncached.

This is stacked on #122.


judgearena/inference.py
```diff
@@ -23,43 +24,103 @@
...
+    revision or an explicit ``chat_template``.
+    """
+    input_mode = provider_input_mode(provider)
+    if input_mode is None:
         return None
-    return {
+
+    descriptor_kwargs = resolved_kwargs.copy()
+    if provider == "VLLM":
+        descriptor_kwargs["temperature"] = (
```
kargibora, line 97: This is confusing. First if we want to change the default temperature this may be missed. We can instead make this a constant OR just trust that `descriptor_kwargs` will be a resolves task (so all of the parameters are set). Then we can use it instead (which has the default temperature stored).

Also we can also use
```py
descriptor_kwargs.get("temperature", DEFAULT_TEMPERATURE)
```
this looks cleaner.  

So we should resolve thsi in `_resolve_model_config()` and not here in my opinion

judgearena/models.py
```diff
@@ -893,6 +894,61 @@ def _route_sampling_params(
     return engine_kwargs
 
 
+_VLLM_ONLY_KWARGS = (
```
kargibora, line 905: I am not sure with this design. It is very hard-coded. Do we really need to check this?
(Even if this exist or introduced before)

judgearena/inference.py
kargibora: Also we use this value both in `inference.py` and `ChatVLLM`

judgearena/inference.py
```diff
@@ -23,43 +24,103 @@
...
+    "gpu_memory_utilization",
+    "tensor_parallel_size",
+}
 
+InputMode = Literal["chat", "text"]
 
-def canonicalize_chat_input(input_item: Any) -> str:
-    """Serialize a logical model input for content-addressed cache lookup."""
+
+def provider_input_mode(provider: str) -> InputMode | None:
```
kargibora, line 46: Can't we have a single PROVIDERS that returns it modes?

```PY
_PROVIDERS ={"OpenRouter" : "chat"...}
```

judgearena/inference.py
ErlisLushtaku: Agreed, both vLLM defaults now live in shared constants and are used by the descriptor and `ChatVLLM`.

judgearena/models.py
ErlisLushtaku: I agree it seems not very general, but I kept this because shared `engine_kwargs` can include vLLM-only settings for hosted providers, which would otherwise receive them as request parameters. If you have any different idea how to address this, let me know.

judgearena/inference.py
ErlisLushtaku: Agreed, this is a single provider-to-input-mode map now.

judgearena/models.py
kargibora: Yes I agree. Perhaps this is a matter of another PR.

kargibora (approved):

PR #124, opened 2026-09-09
Cache (4/5): Adopt inference caching in task runners

ErlisLushtaku:
# Description

Adopts the inference cache in the task runners through `--run.store_root`.

- Pairwise, meta-eval, MT-Bench, and Elo generation and judging share the `do_inference` cache.
- Elo temperature calibration uses the same judgement cache.
- Cache rows get instruction, model pair, and orientation metadata at the direct inference call.
- Dataset-provided completions stay as direct inputs when they do not run inference.
- Removes `ignore_cache`, `cache_function_dataframe`, and the old cache tokens.

This is stacked on #123.


judgearena/benchmarks/elo/runner.py
```diff
@@ -121,48 +126,13 @@ def run_elo(cfg: "RunConfig", task: ResolvedTaskSpec | None = None) -> dict:
...
-    gen_fun = partial(
-        generate_instructions,
+    completions_df = generate_instructions(
+        instructions=instructions,
+        model=cfg.model.name,
         truncate_input_chars=cfg.generation.truncate_all_input_chars,
         use_tqdm=use_tqdm,
+        inference_cache=build_completion_cache(cfg),
         **extra_kwargs,
-    )
```
kargibora, line 129: Nice that we get rid of this!

kargibora (commented):
Overall this PR looks quite clear. It wires all the things that have been implemented and removes the legacy code. It is good that we got rid of many lines that were making the code harder to parse!

kargibora (approved):

kargibora (commented):
I also did not understand the reason of `cache_metadata`.  Do we need it in , `do_inference(),  annotate_battles() and judge_and_parse_prefs()`?

ErlisLushtaku:
@kargibora Yes, it needs to reach all three in the current flow. It is not part of the cache key, but rather the info that doesn't necessarily affect determine the output e.g. `instruction_id`, `model_a`, `orientation` etc. `do_inference` uses it only when storing a miss, `annotate_battles` keeps it aligned with the rendered prompts, and `judge_and_parse_prefs` swaps `model_a`/`model_b` and orientation for the reversed prompt. I renamed it to `cache_row_metadata` to make this distinction clearer.

We could avoid passing a separate argument by creating an object such as:
`InferenceInput(payload=prompt, cache_row_metadata=metadata)`

What do you think?

kargibora:
Thanks for making it clear! I think that can work

PR #127, opened 2026-09-09
Cache (5/5): Synchronize inference caches through Hugging Face

ErlisLushtaku:
# Description

Adds an explicit CLI for sharing inference caches through a Hugging Face dataset repository.

- Fetches and pushes folders filtered by role, task and optional model.
- Validates descriptors and merges duplicate input hashes by the latest `pushed_at`.
- Uploads `metadata.json` and the SQLite database in one commit.
- Retries parent-commit conflicts after fetching and merging the newer remote rows.
- Documents local caching and the fetch/push commands in the README.

Synchronization remains separate from benchmark execution. The Parquet store, migration scripts, and cache browser from #71 are intentionally left out.

This is stacked on #124.


judgearena/cache/hf.py
```diff
@@ -0,0 +1,295 @@
+"""Filtered Hugging Face synchronization for inference-cache folders."""
```
kargibora, line 1: I am not sure whether we shoudl expose this in `judgearena/` perhaps it is better to put this to `judgearena/utils` or `judgearena/cache/...` (similarly other materials into this as well as we introduce more cache specific things, it makes more sense to create seperate package for it.

kargibora: (If we want to expose it as cache it makes sense to me that this stays under judgearena but still wanted to discuss)

kargibora (approved):
Looks good to me, just a several comments

judgearena/cache/hf.py
ErlisLushtaku: Agreed, moved the cache CLI to `judgearena/cache/hf.py` and updated the entry point to `judgearena.cache.hf:cli`. The other cache-specific modules now live under the same package.

